PR #88, opened 2015-01-01
Refactored SLRE command line utility

cpq:


Makefile
```diff
@@ -21,10 +21,6 @@ v: unit_test
 #	valgrind -q --leak-check=full ./v7 tests/run_tests.js
 #	gcov -a unit_test.c
 
-xrun: unit_test
-	$(CC) -W -Wall -I. -I./src src/tokenizer.c -DTEST_RUN -DV7_EXPOSE_PRIVATE -o t
-	./t
-
 run:
 	@$(MAKE) -C tests coverage
```
mmikulicic: Should we have a rule that actually runs this code or do you plan to put it in the slre repo?

v7.c
```diff
@@ -10930,173 +10949,110 @@ static unsigned char get_flags(const char *ch) {
...
+    err_code = process_line(pr, flags, str, cap_no, replace, verbose);
+  } else if (file_name != NULL) {
+    FILE *fp = strcmp(file_name, "-") == 0 ? stdin : fopen(file_name, "rb");
+    char line[20 * 1024];
+    if (fp == NULL) {
+      fprintf(stderr, "Cannot open %s: %s\n", file_name, strerror(errno));
+      exit(1);
+    } else {
+      while (fgets(line, sizeof(line), fp) != NULL) {
+        err_code = process_line(pr, flags, line, cap_no, replace, verbose);
```
mmikulicic: This will return an error status if the last line doesn't match.
Probably it should return error only if there are no matching lines.

src/slre.c
```diff
@@ -1184,7 +1184,7 @@ static void re_newthread(struct slre_thread *t, struct slre_instruction *pc,
   if (!(thr = 0)) continue
 
 static unsigned char re_match(struct slre_instruction *pc, const char *start,
-                              const char *bol, unsigned char flags,
+                              const char *bol, unsigned flags,
```
mmikulicic: Throughout the rest of our code base we explicitly use `unsigned int`. AFAIK `unsigned` and `unsigned int` are synonyms.

Makefile
cpq: Reverted that change, it doesn't belong to this PR.
TBD in separate PR.
This is an old, outdated tokenizer "test" that should be removed from the makefile and tokenizer.c.

v7.c
```diff
@@ -10886,37 +10885,57 @@ int slre_replace(struct slre_loot *loot, const char *src, const char *rstr,
...
-    t++;
-  } while (--i);
-  *d = '\0';
-  return osz;
+
+  return res;
+}
+
+
+#ifdef SLRE_TEST
```
mmikulicic: I commented on the makefile because it was relevant there, but it might as well be relevant here :-)

Should we have a makefile rule that actually runs this code or do you plan to put it in the slre repo?

For lazymen like me who don't want to type `gcc -o /tmp/slre src/slre.c src/utf.c -DSLRE_TEST` :-)

src/slre.c
cpq: Changed to `unsigned int`

v7.c
cpq: The plan is to move this into SLRE repo.
The main purpose of this binary is not to provide an actual grep-like functionality,
but to be a debugging tool for the regex engine. The tool should be able to dump
captures, and perhaps generated bytecode too.
As we plan to use ECMA tests for v7, I am not sure on whether we need to have a hand-crafted tests for SLRE as well.

v7.c
cpq: Changed.

PR #90, opened 2015-01-02
Optimize `cmp` function.

vstakhov:
Before:
% time ./v7 fib.js
[0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597,
2584, 4181, 6765, 10946, 17711, 28657, 46368, 75025, 121393, 196418,
317811, 514229, 832040]
7049123
./v7 fib.js  71,47s user 0,13s system 99% cpu 1:11,85 total

After:
% time ./v7 fib.js
[0, 1, 1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, 610, 987, 1597,
2584, 4181, 6765, 10946, 17711, 28657, 46368, 75025, 121393, 196418,
317811, 514229, 832040]
7049123
./v7 fib.js  42,39s user 0,08s system 99% cpu 42,573 total

Compiled with gcc 4.8 using

% gcc -O3 -g -DNDEBUG -DV7_EXE v7.c -o v7 -lm

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -269,31 +269,15 @@ V7_PRIVATE struct v7_val v7_str_to_val(const char *buf) {
 }
 
 V7_PRIVATE int cmp(const struct v7_val *a, const struct v7_val *b) {
-  int res;
   double an, bn;
   const struct v7_string *as, *bs;
-  struct v7_val ta = MKOBJ(0), tb = MKOBJ(0);
 
   if (a == NULL || b == NULL) return 1;
+  if (a->type != b->type) return 1;
+
   if ((a->type == V7_TYPE_UNDEF || a->type == V7_TYPE_NULL) &&
       (b->type == V7_TYPE_UNDEF || b->type == V7_TYPE_NULL))
     return 0;
-
-  if (is_num(a) && is_num(b)) {
-    v7_init_num(&ta, a->v.num);
-    v7_init_num(&tb, b->v.num);
-    a = &ta;
-    b = &tb;
-  }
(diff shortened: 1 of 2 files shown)


cpq:
v7.c is a generated file, amalgamated from src/*.c
Please update the source, src/util.c then do `make` in the root level dir to generate v7.c

vstakhov:
Done. However, v7 is still terribly slow. I have the updated profile for this simple test. Previously, the function `cmp` ate about 40% of time. At the moment, there are no such specific functions, however the overall performance is still unacceptable. 30 Fibonacci numbers means 7M of recursive calls approximately: `v8` performs that in likely 0.15 seconds, `duktape` - in 4.5 seconds (which is awful as well) and v7 spends about 40 seconds now (but it is definitely better than 72 seconds without this patch).

cpq:
I still don't see changes in `util.c`. Did you do `git commit -a --amend` followed by `git push -f` ?

cpq:
Ah, got it. There was a second commit.
Merging.
W.r.t. performance - agree with the point. We're in the middle of major core refactoring, which is going to improve things a lot. The 4 loosely coupled components of the engines are: tokenizer, parser, ast, VM - and first three are more or less done. We can parse all ECMA tests and libs like jquery, and generate AST for those. Work is in progress on VM part.
Note that right now an "old" components coexists with "new" ones: like, old "parser.c" with new "aparser.c".

PR #91, opened 2015-01-06
Moved mbuf into a separate module

cpq:


src/main.c
```diff
@@ -37,15 +37,15 @@ static char *read_file(const char *path, size_t *size) {
 static void dump_ast(const char *code, int binary) {
   struct ast ast;
 
-  ast_init(&ast, 0);
+  mbuf_init(&ast.mbuf, 0);
```
mmikulicic: Since we keep the `struct ast` (mostly for self documenting purposes, but it might be useful to keep more AST related stuff) wouldn't it make sense to keep the `ast_init` function so that we don't have to locate all invocations of `mbuf_init` that initialize a `struct ast::buf` value?

cpq: Makes sense. That will keep AST implementation private to AST.
I've changed ast_trim() to ast_optimize().
PTAL.

tests/unit_test.c
```diff
@@ -872,7 +872,7 @@ static const char *test_aparser(void) {
   char want_ast[102400];
   char *next_want_ast = want_ast - 1;
   size_t want_ast_len;
-  ast_init(&a, 0);
+  mbuf_init(&a.mbuf, 0);
```
mmikulicic: please also revert tests

mmikulicic:
LGTM modulo reverting unit_test

cpq:
unit test reverted, merging.

PR #95, opened 2015-01-08
Add var threading and function declarations

mmikulicic:
Function objects now are created as special function objects
and tagged as such while being treated as objects during get/set.
The json rendering for functions shows the arguments and hoisted
var decls for now mostly as a debugging aid and will likely be
removed once we can test function invocation.

----

The code that will actually build the stack frame upon function entry
will be taken/inspired from `v7_to_json`, keep this in mind while
reviewing the code.


mmikulicic:
Updated with addressed comments, and a few more tests (coverage driven) and fixes (exec allocates ast mbuf struct on heap instead on stack, so that function values can be used after exec returns)

PR #96, opened 2015-01-09
String API

cpq:


src/string.c
```diff
@@ -5,6 +5,101 @@
...
+    tag |= V7_TAG_OWNED;
+  } else {
+    embed_string(m, m->len, (char *) &p, sizeof(p));
+  }
+
+  return v7_pointer_to_value((void *) oft) | tag;
+}
+
+/* Get a pointer to string and string length */
+static char *s_get_str(struct v7 *v7, val_t v, size_t *sizep) {
```
mmikulicic: We didn't converge yet on a new naming convention. Here his function has "s_" prefix (according the your proposal in https://cesanta.slack.com/archives/general/p1420624132000004), but `embed_string` has none, and `to_string` either.

cpq: That's right. I suggested internal API either not prefixed, or prefixed with module prefix. The functions you've mentioned not prefixed.
We can agree to have all internal API functions prefixed.

src/string.c
```diff
@@ -5,6 +5,101 @@
...
+  int k = calc_llen(n);           /* Calculate how many bytes length takes */
+  mbuf_insert(m, oft, NULL, k);   /* Allocate  buffer for length */
+  encode_varint(n, (unsigned char *) m->buf + oft);  /* Write length */
+  mbuf_insert(m, oft + k, p, n);  /* Copy the string itself */
+}
+
+/* Create a string */
+V7_PRIVATE val_t to_str(struct v7 *v7, const char *p, size_t len, int own) {
+  struct mbuf *m = own ? &v7->owned_strings : &v7->foreign_strings;
+  size_t oft = m->len;
```
mmikulicic: What does `oft` stand for?

PR #97, opened 2015-01-10
Implement function call

mmikulicic:
Note: call frames are not yet linked together so each function can only
access it's own variables.

DIFFBASE=#95


cpq:
Very neat. LGTM, pls see my  comments  (minor stuff).

mmikulicic:
PTAL new `ast_get_inlined_data` and it's uses

PR #98, opened 2015-01-11
Add C functions support

cpq:
DIFFBASE=96

This is a sketch for the C/JavaScript glue API.
Added print() function to the global object for debugging.
Switched command line utility from v7_exec to v7_exec_2 - that is the beginning of wiping out old functionality.
Imlemented VAR_DECL

### DIFF
--- a/src/api.c
+++ b/src/api.c
@@ -5,6 +5,20 @@
 
 #include "internal.h"
 
+/* TODO(lsm): remove this when init_stdlib() is upgraded */
+V7_PRIVATE v7_val_t Std_print_2(struct v7 *v7, val_t args) {
+  char *p;
+  int i, num_args = v7_array_length(v7, args);
+  for (i = 0; i < num_args; i++) {
+    p = debug_json(v7, v7_array_at(v7, args, i));
+    printf("%s", p);
+    free(p);
+  }
+  putchar('\n');
+
+  return v7_create_value(v7, V7_TYPE_NULL);
+}
+
 struct v7 *v7_create(void) {
   static int prototypes_initialized = 0;
   struct v7 *v7 = NULL;
@@ -23,12 +37,16 @@ struct v7 *v7_create(void) {
      * Ensure the first call to v7_create_value will use a null proto:
      * {}.__proto__.__proto__ == null
      */
-    v7->object_prototype = NULL;
     v7->object_prototype = val_to_object(
         v7_create_value(v7, V7_TYPE_GENERIC_OBJECT));
     v7->array_prototype = val_to_object(
         v7_create_value(v7, V7_TYPE_GENERIC_OBJECT));
     v7->global_object = v7_create_value(v7, V7_TYPE_GENERIC_OBJECT);
+
+    /* TODO(lsm): remove this when init_stdlib() is upgraded */
+    v7_set_property_value(v7, v7->global_object, "print", 5, 0,
+                          v7_create_value(v7, V7_TYPE_CFUNCTION_OBJECT,
+                                          Std_print_2));
   }
 
   return v7;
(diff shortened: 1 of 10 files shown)


src/api.c
```diff
@@ -23,12 +37,16 @@ struct v7 *v7_create(void) {
...
      */
-    v7->object_prototype = NULL;
     v7->object_prototype = val_to_object(
         v7_create_value(v7, V7_TYPE_GENERIC_OBJECT));
     v7->array_prototype = val_to_object(
         v7_create_value(v7, V7_TYPE_GENERIC_OBJECT));
     v7->global_object = v7_create_value(v7, V7_TYPE_GENERIC_OBJECT);
+
+    /* TODO(lsm): remove this when init_stdlib() is upgraded */
+    v7_set_property_value(v7, v7->global_object, "print", 5, 0,
```
mmikulicic: Side note: we should probably rename `v7_set_property_value` to `v7_set_property`

src/interpreter.c
```diff
@@ -244,7 +259,19 @@ static val_t i_eval_call(struct v7 *v7, struct ast *a, ast_off_t *pos, val_t sco
...
+
+  if (v7_is_cfunction(v1)) {
+    char buf[20];
+    int n, i;
+    val_t args = v7_create_value(v7, V7_TYPE_ARRAY_OBJECT);
+    for (i = 0; *pos < end; i++) {
+      res = i_eval_expr(v7, a, pos, scope);
+      v7_array_append(v7, args, res);
+      n = snprintf(buf, sizeof(buf), "%d", i);
+      v7_set_property_value(v7, args, buf, n, 0, res);
```
mmikulicic: Doesn't `v7_array_append` already set the value?

src/vm.c
```diff
@@ -467,22 +483,40 @@ V7_PRIVATE val_t v7_property_value(struct v7_property *p) {
   return p->value;
 }
 
-V7_PRIVATE val_t v7_array_length(struct v7 *v7, val_t v) {
-  struct v7_property *prop;
+V7_PRIVATE long v7_array_length(struct v7 *v7, val_t v) {
```
mmikulicic: Please also fix unit tests that check for array length treating it as `val_t`, e.g.:
``` ASSERT(check_value(v7, v7_array_length(v7, v), "3"));```

File: v7.c [hunk 4]
```diff
@@ -9763,9 +9823,23 @@ static val_t i_eval_expr(struct v7 *v7, struct ast *a, ast_off_t *pos,
       }
     case AST_CALL:
       return i_eval_call(v7, a, pos, scope);
+    case AST_VAR:
+      end = ast_get_skip(a, *pos, AST_END_SKIP);
+      ast_move_to_children(a, pos);
+      while (*pos < end) {
+        size_t name_len;
+        const char *name;
+        enum ast_tag tag = ast_fetch_tag(a, pos);
```
mmikulicic: No need to shadow the tag variable:

```assert((tag = ast_fetch_tag(a, pos)) == AST_VAR_DECL);```

src/api.c
cpq: Yes. And get rid of the other function (vararg set_property).
Also I think we should get rid of generic v7_create_object() and export per-type creation functions: there is a high chance of screwing up vararg.

src/interpreter.c
cpq: oops :-)
overlook. removing v7_array_append() call from here, cause it is slower, and we can have C function calls done in loops, so we'd rather be fast.

src/vm.c
cpq: Yep done!

v7.c
cpq: Done.
Aslo, I've moved all ast_fetch_tag(a, pos) calls out of assert() invocation (with -DNDEBUG, those  important ast_fetch_tag(a, pos)  which shift position, will be optimized out).

PR #99, opened 2015-01-12
Traverse prototype chain

mmikulicic:
Implement function invokation scope as chain
of frames chained via the __proto__ field, so that
variables from outer lexical scopes are looked up
via the same mechanism as object property lookup.

Fix prototype name comparison.

Allow ECMA test sandboxing so that unit test doesn't crash
when we enable a codepath that triggers an interpreter bug.
(disabled by default, enable in tests/Makefile)


mmikulicic:
Note that the traversal algorithm is incorrect, and will be fixed when we start supporting object prototypes, but for now it's the quickest way to get function evaluation and basic object access working

cpq:
Ack, thx!

PR #112, opened 2015-01-13
Using simple stubs for ECMA driver

cpq:


mmikulicic:
Already merged through #114

PR #145, opened 2015-01-14
Split v7_create_value to type-specific functions

cpq:
Added $(realpath) to makefiles to let IDE know the exact file location to jump to compilation errors.
Fixed warning in interpreter.c
Exported creation functions to v7.h

### DIFF
--- a/src/interpreter.c
+++ b/src/interpreter.c
@@ -150,8 +150,8 @@ static val_t i_eval_expr(struct v7 *v7, struct ast *a, ast_off_t *pos,
   switch (tag) {
     case AST_NEGATIVE:
     case AST_POSITIVE:
-      return v7_create_value(v7, V7_TYPE_NUMBER, i_num_unary_op(
-          v7, tag, i_as_num(v7, i_eval_expr(v7, a, pos, scope))));
+      return v7_create_number(i_num_unary_op(v7, tag, i_as_num(v7,
+                              i_eval_expr(v7, a, pos, scope))));
     case AST_ADD:
       v1 = i_eval_expr(v7, a, pos, scope);
       v2 = i_eval_expr(v7, a, pos, scope);
@@ -160,10 +160,10 @@ static val_t i_eval_expr(struct v7 *v7, struct ast *a, ast_off_t *pos,
         v7_stringify_value(v7, v1, buf, sizeof(buf));
         v7_stringify_value(v7, v2, buf + strlen(buf),
                            sizeof(buf) - strlen(buf));
-        return v7_create_value(v7, V7_TYPE_STRING, buf, strlen(buf));
+        return v7_create_string(v7, buf, strlen(buf), 1);
       }
-      return v7_create_value(v7, V7_TYPE_NUMBER, i_num_bin_op(v7, tag,
-                             i_as_num(v7, v1), i_as_num(v7, v2)));
+      return v7_create_number(i_num_bin_op(v7, tag, i_as_num(v7, v1),
+                              i_as_num(v7, v2)));
     case AST_SUB:
     case AST_REM:
     case AST_MUL:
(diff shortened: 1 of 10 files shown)


src/interpreter.c
```diff
@@ -325,26 +325,26 @@ static val_t i_eval_expr(struct v7 *v7, struct ast *a, ast_off_t *pos,
...
       return res;
     case AST_TRUE:
-      return v7_create_value(v7, V7_TYPE_BOOLEAN, 1);
+      return v7_create_boolean(1);
     case AST_FALSE:
-      return v7_create_value(v7, V7_TYPE_BOOLEAN, 0);
+      return v7_create_boolean(0);
     case AST_NULL:
-      return v7_create_value(v7, V7_TYPE_NULL);
+      return v7_create_null();
```
mmikulicic: return V7_NULL;

src/interpreter.c
```diff
@@ -325,26 +325,26 @@ static val_t i_eval_expr(struct v7 *v7, struct ast *a, ast_off_t *pos,
...
     case AST_NOP:
     case AST_UNDEFINED:
-      return v7_create_value(v7, V7_TYPE_UNDEFINED);
+      return v7_create_undefined();
```
mmikulicic: return V7_UNDEFINED

v7.h
```diff
@@ -23,7 +23,10 @@
 extern "C" {
 #endif /* __cplusplus */
 
+#include <stddef.h>   /* For size_t */
+
 #define V7_VERSION "1.0"
+#define V7_GENERIC_OBJECT_PROTOTYPE 0     /* For v7_create_object() */
```
mmikulicic: This is an valid `val_t` for double `+0`.

Why not use V7_NULL for NULL prototype and V7_UNDEFINED for default proto?

cpq: Cause V7_GENERIC_OBJECT_PROTOTYPE is defined in public v7.h
In order to expose V7_NULL, we need to expose quite a bit of stuff to the user - our packing mechanism.

mmikulicic: `(v7_create_null())` ?

cpq: SGTM, updated

src/interpreter.c
```diff
@@ -547,7 +547,7 @@ static val_t i_eval_call(struct v7 *v7, struct ast *a, ast_off_t *pos, val_t sco
   ast_skip_tree(func->ast, &fpos);
   fargs = fpos;
 
-  frame = v7_create_value(v7, V7_TYPE_GENERIC_OBJECT);
+  frame = v7_create_object(v7, V7_NULL);
```
mmikulicic:
```
frame = v7_create_object(v7, v7_object_to_val(func->scope));
```

v7.h
mmikulicic: I'm sorry, I meant `(v7_create_undefined())`

Also `Object.create(null)` is a valid way to create a new object with a null prototype.

src/vm.c
```diff
@@ -159,71 +159,54 @@ static val_t v7_create_object(struct v7 *v7, struct v7_object *proto) {
   }
   (void) v7;
   o->properties = NULL;
-  o->prototype = proto;
+  o->prototype = prototype == V7_UNDEFINED ? NULL : val_to_pointer(prototype);
```
mmikulicic: let's use `V7_NULL` here for `NULL`, since that's the value that will be passed by the `Obj_create` primitive (which accepts JS `null` as a valid value, while it rejects `undefined`. V8 says:

```
TypeError: Object prototype may only be an Object or null
```

Spidermonkey:
```
TypeError: Expected argument of type object, but instead had type undefined
```

mmikulicic: Or let's use `val_to_object` here and convert a V7_NULL into a NULL pointer in `val_to_object` (like we do the other way around in `v7_object_to_value`

(btw, at some point we should cleanup the `v7_xxx_to_value` names to `(y_)?xxx_to_val`

mmikulicic:
LGTM!

PR #147, opened 2015-01-14
Add Boolean ctor

cpq:
This PR also changes `v7_cfunction_t` signature to the agreed one

### DIFF
--- a/v7.c
+++ b/v7.c
@@ -35,7 +35,7 @@ struct v7_val; /* Opaque structure. Holds V7 value, which has v7_type type. */
 #include <inttypes.h>
 typedef uint64_t v7_val_t;
 
-typedef v7_val_t (*v7_cfunction_t)(struct v7 *, v7_val_t args);
+typedef v7_val_t (*v7_cfunction_t)(struct v7 *, v7_val_t, v7_val_t);
 
 struct v7 *v7_create(void);     /* Creates and initializes V7 engine */
 void v7_destroy(struct v7 *);   /* Cleanes up and deallocates V7 engine */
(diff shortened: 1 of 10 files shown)


v7.c
```diff
@@ -757,6 +757,9 @@ typedef uint64_t val_t;
 #define V7_TAG_STRING_F  ((uint64_t) 0xFFF8 << 48)  /* Foreign string */
 #define V7_TAG_FUNCTION  ((uint64_t) 0xFFF7 << 48)  /* JavaScript function */
 #define V7_TAG_CFUNCTION ((uint64_t) 0xFFF6 << 48)  /* C function */
+#define V7_TAG_O_BOOLEAN ((uint64_t) 0xFFF5 << 48)  /* Boolean object */
```
mmikulicic: I'm not sure we should waste tag space for this objects, which are just plain objects with a hidden property.

```
Object.getPrototypeOf(new Boolean(1)) === Boolean.prototype
```

The only thing this object has to know is to covert to number so that it can behave correctly in arithmetics (and hence in comparison operators):

```
> new Boolean(0) == 0
true
> new Boolean(0) - 1
-1
> new Boolean(1) - 1
0
```

In fact, in V8 if you change the prototype of a Boolean object, it's no longer a boolean:
```
> b = new Boolean(1)
{}
> b == 1
true
> b.__proto__ = {}
{}
> b == 1
false
```

So, if we tag the value, we should untag all the values pointing to it, whenever it's prototype changes.

I'd rather do the same thing we did for arrays: use the instance's `prototype` to decide what to do when doing things like converting to number.

cpq: Done, PTAL

PR #156, opened 2015-01-15
Implement eval()

cpq:
Unescape strings when putting them into AST.
Disable bunch of ECMA tests that segfault the VM.


src/vm.c
```diff
@@ -663,6 +684,23 @@ V7_PRIVATE int is_prototype_of(val_t o, val_t p) {
   return 0;
 }
 
+static val_t Std_eval(struct v7 *v7, val_t t, val_t args) {
+  val_t res = V7_UNDEFINED, arg = v7_array_at(v7, args, 0);
+  (void) t;
+  if (arg != V7_UNDEFINED) {
+    char buf[100], *p;
+    p = v7_to_json(v7, arg, buf, sizeof(buf));
```
mmikulicic: You could use `v7_stringify_value` for that. (Which I guess needs some more love)

src/vm.c
```diff
@@ -663,6 +684,23 @@ V7_PRIVATE int is_prototype_of(val_t o, val_t p) {
...
+    if (p[0] == '"') {
+      p[0] = p[strlen(p) - 1] = ' ';
+    }
+    res = v7_exec(v7, p);
```
mmikulicic: As discussed in chat, reentrant execution of `v7_exec` needs to be fixed by saving the content of the jump bufs in v7.

mmikulicic:
LGTM, modulo comments

src/vm.c
cpq: I'd stick with v7_to_json() cause it doesn't have a restrictio on buffer
size

src/vm.c
cpq: Done, PTAL

src/vm.c
mmikulicic: SGTM

src/interpreter.c
```diff
@@ -953,6 +958,11 @@ V7_PRIVATE val_t v7_exec(struct v7 *v7, const char* src) {
 #if 0
   fprintf(stderr, "Eval res: %s .\n", debug);
 #endif
+
+  /* Restore exception environments */
+  memcpy(&v7->jmp_buf, &saved_jmp_buf, sizeof(saved_jmp_buf));
```
mmikulicic: this function returns in other places, for example an exception is thrown:

```
function testcase() {
    eval("throw 1");
    return prop;
}
testcase();
```

`zsh: abort      ./v7 /tmp/test3.js`

I think a `goto cleanup` is the cleanest way (I'm using this in another branch for restoring `v7->this_object` for `v7_exec_with`)

cpq: done

PR #167, opened 2015-01-20
Add load_json_config example

### LINKED ISSUE #121: Add example for loading JSON config

cpq:
Export v7_to_* functions
Fix #121

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -1,0 +1,14 @@
+# Copyright (c) 2015 Cesanta Software
+# All rights reserved
+
+SUBDIRS = $(sort $(dir $(wildcard */)))
+
+.PHONY: $(SUBDIRS)
+
+all: $(SUBDIRS)
+
+$(SUBDIRS):
+	@$(MAKE) -C $@
+
+clean:
+	for d in $(SUBDIRS) ; do $(MAKE) -C $$d clean ; done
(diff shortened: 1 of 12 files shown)


examples/Makefile
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2015 Cesanta Software
+# All rights reserved
+
+SUBDIRS = $(sort $(dir $(wildcard */)))
+X = $(SUBDIRS)
```
mmikulicic: garbage?

cpq: Removed.

v7.h
```diff
@@ -52,7 +52,19 @@ v7_val_t v7_create_null(void);
...
+int v7_to_boolean(v7_val_t);
+double v7_to_double(v7_val_t);
+v7_cfunction_t v7_to_cfunction(v7_val_t);
+const char *v7_to_string(struct v7 *, v7_val_t *, size_t *);
+
 v7_val_t v7_get_global_object(struct v7 *);
+
+/*
+ * TODO(lsm): rename either to v7_get_property(), or
+ * rename v7_set_property() to v7_set()
```
mmikulicic: IMHO the user sometimes needs just to update a property value without actually thinking about the property in terms of it's attributes.

Hence, I see a `v7_set` function which will set the value of a property, by keeping its attributes unchanged (or create one with default attributes) and a `v7_define_property` which will actually allow the user to set the property attributes analogous to the JS `Object.defineProperty` function.

Since `v7_get_property` returns a pointer to an VM internal structure, we could offer a C helper analogous to the `Object.getOwnPropertyDescriptor`.

PR #176, opened 2015-01-20
Fix empty statement value

mmikulicic:
Should be `undefined` not `null`.


cpq:
failed?

mmikulicic:
yep, didn't update a manual unit test

mmikulicic:
Fixed ptal

PR #180, opened 2015-01-21
Complete Math object implementation

### LINKED ISSUE #129: Implement Math interface

cpq:
Fix #129


src/math.c
```diff
@@ -5,6 +5,38 @@
 
 #include "internal.h"
 
+static val_t one_arg(struct v7 *v7, val_t args, double (*f)(double)) {
```
mmikulicic: perhaps it's better to prefix these as `m_one_arg`

mmikulicic:
LGTM, modulo minor comment

src/math.c
cpq: Done

PR #185, opened 2015-01-22
String object partial implementation

cpq:


src/string.c
```diff
@@ -5,46 +5,55 @@
...
-    runes[n] = _conv_to_int(cfa->v7, cfa->args[n]);
-    blen += runelen(runes[n]);
+V7_PRIVATE val_t Str_charCodeAt(struct v7 *v7, val_t this_obj, val_t args) {
+  size_t i, n;
+  const char *p = v7_to_string(v7, &this_obj, &n), *end = p + n;
+  val_t res = v7_create_number(NAN), arg = v7_array_at(v7, args, 0);
+
+  if (v7_is_double(arg) && v7_to_double(arg) >= 0 && v7_is_string(this_obj)) {
+    Rune r;
+    for (i = 0; i < (size_t) v7_to_double(arg) && p < end; i++) {
```
mmikulicic: please store the result of `v7_to_double` in case the compiler doesn't inline it

mmikulicic:
LGTM, modulo minor comment

src/string.c
cpq: Done

PR #186, opened 2015-01-22
Add support for getters

mmikulicic:
Setters TODO.
Changed AST format to make it easier to interpret
while making the AST for getters marginally bigger,
which is not a big deal since they are rare constructs.

Refactored function calls and added a `v7_apply` function
to call any V7 function from C.


cpq:
Minor comment.
LGTM.

PR #188, opened 2015-01-23
Fix JSON config example

cpq:
Make `v7_exec_file` fail if file could not be opened.


src/interpreter.c
```diff
@@ -1122,7 +1122,7 @@ enum v7_err v7_exec_file(struct v7 *v7, val_t *res, const char *path) {
   FILE *fp;
   char *p;
   long file_size;
-  enum v7_err err = V7_OK;
+  enum v7_err err = V7_EXEC_EXCEPTION;
```
mmikulicic: shouldn't we also set something in the `val`, e.g. an exception object ?

cpq: Added exception.

PR #190, opened 2015-01-23
Hardcode String/Number prototype lookup

cpq:
DIFFBASE=189
Fix String.charCodeAt().


mmikulicic:
LGTM, needs rebase since parent commit has been amended

PR #197, opened 2015-01-27
Move valueOf from String to Object

cpq:
Added `i_value_of()` function that extracts primitive value from V7_TYPE_STRING_OBJECT, V7_TYPE_NUMER_OBJECT, V7_TYPE_BOOLEAN_OBJECT  objects.
When such objects are subjects to arithmetic / logical operations, a HIDDEN property (i.e. valueOf) must be extracted instead.
Re-enabled bunch of tests.


src/object.c
```diff
@@ -187,6 +199,7 @@ V7_PRIVATE void init_object(struct v7 *v7) {
   object = v7_get(v7, v7->global_object, "Object", 6);
   v7_set(v7, object, "prototype", 9, v7->object_prototype);
 
+  object = v7->object_prototype;
   set_cfunc_prop(v7, object, "getPrototypeOf", Obj_getPrototypeOf);
   set_cfunc_prop(v7, object, "isPrototypeOf", Obj_isPrototypeOf);
```
mmikulicic: `isPrototypeOf` should be set on `v7->object_prototype` (like `hasOwnProperty`)

src/object.c
```diff
@@ -187,6 +199,7 @@ V7_PRIVATE void init_object(struct v7 *v7) {
   object = v7_get(v7, v7->global_object, "Object", 6);
   v7_set(v7, object, "prototype", 9, v7->object_prototype);
 
+  object = v7->object_prototype;
```
mmikulicic: This is incorrect. The following properties are defined on `Object` not on `Object.prototype` (except `isPrototypeOf` see below).

mmikulicic:
LGTM, modulo comment

src/object.c
cpq: Done

src/object.c
cpq: Removed.

mmikulicic: Slipped through, fix in #200 
(`({}).keys` should yield `undefined`)

PR #199, opened 2015-01-27
Implement {String,Array}.prototype.length

cpq:
This fix reduces coverage, because SUDDENLY loops in tests start to work!  (e.g. `for (i =0; i < arr.length; i++`)


cpq:
PTAL

PR #211, opened 2015-01-28
Fix Number conversion and arithmetics

mmikulicic:
* Fix infinite recursion when throwing an exception while
  constructing an exception object to throw.

* Fix property attributes for important properties
  and move constants to the `Number` object and not in the number prototype.

* Define `MAX_VALUE` and `MIN_VALUE` according to the spec.

* Fix rendering of Infinity numbers and use a sprintf precision
  that matches more closely the rules in:
  http://www.ecma-international.org/ecma-262/5.1/#sec-9.8.1
  so that more test are excercised (although it's not perfect).

* Hack ordering of `Object.keys()`, in order to pass some tests were
  unlocked by other fixes.


cpq:
Pls fix tests

mmikulicic:
PTAL

PR #223, opened 2015-02-05
Fix Number ctor

cpq:
Note: there is a regression.

### DIFF
--- a/src/interpreter.c
+++ b/src/interpreter.c
@@ -703,7 +703,7 @@ static val_t i_eval_expr(struct v7 *v7, struct ast *a, ast_off_t *pos,
     case AST_INSTANCEOF:
       v1 = i_eval_expr(v7, a, pos, scope);
       v2 = i_eval_expr(v7, a, pos, scope);
-      if (!v7_is_function(v2)) {
+      if (!v7_is_function(v2) && !v7_is_cfunction(i_value_of(v7, v2))) {
         throw_exception(v7, "TypeError",
                         "Expecting a function in instanceof check");
       }
@@ -824,6 +824,9 @@ static val_t i_eval_call(struct v7 *v7, struct ast *a, ast_off_t *pos,
   end = ast_get_skip(a, *pos, AST_END_SKIP);
   ast_move_to_children(a, pos);
   v1 = i_eval_expr(v7, a, pos, scope);
+  if (!v7_is_cfunction(v1) && !v7_is_function(v1)) {
+    v1 = i_value_of(v7, v1);
+  }
 
   if (v7_is_cfunction(v1)) {
     char buf[20];
--- a/src/number.c
+++ b/src/number.c
@@ -6,7 +6,8 @@
 #include "internal.h"
 
 static val_t Number_ctor(struct v7 *v7, val_t this_obj, val_t args) {
-  val_t arg0 = v7_array_at(v7, args, 0);
+  val_t arg0 = v7_array_length(v7, args) <= 0 ?
+    v7_create_number(0.0) : v7_array_at(v7, args, 0);
   val_t res = v7_is_double(arg0) ? arg0 : v7_create_number(i_as_num(v7, arg0));
 
   if (v7_is_object(this_obj) && this_obj != v7->global_object) {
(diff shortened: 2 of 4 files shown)


src/number.c
```diff
@@ -6,11 +6,13 @@
...
 
 static val_t Number_ctor(struct v7 *v7, val_t this_obj, val_t args) {
-  val_t arg0 = v7_array_at(v7, args, 0);
+  val_t arg0 = v7_array_length(v7, args) <= 0 ?
+    v7_create_number(0.0) : v7_array_at(v7, args, 0);
   val_t res = v7_is_double(arg0) ? arg0 : v7_create_number(i_as_num(v7, arg0));
 
   if (v7_is_object(this_obj) && this_obj != v7->global_object) {
     v7_to_object(this_obj)->prototype = v7_to_object(v7->number_prototype);
+    v7_set_property(v7, this_obj, "prototype", 9, 0, v7->number_prototype);
```
mmikulicic: Why this?

cpq: I don't know. Removed.

src/number.c
```diff
@@ -80,18 +82,14 @@ static val_t n_isNaN(struct v7 *v7, val_t this_obj, val_t args) {
...
-          "                        {value:__Number_ctor});"
-          "   return this.__ctor(v)}; "
-          "Number");
-  v7->number_prototype = v7_get(v7, num, "prototype", 9);
-  v7_get_property(num, "prototype", 9)->attributes |= V7_PROPERTY_READ_ONLY;
-
-  v7_set_property(v7, v7->global_object, "__Number_ctor", 13, attrs, ctor);
+  v7_set_property(v7, num, "", 0, V7_PROPERTY_HIDDEN, ctor);
+  v7_set_property(v7, num, "prototype", 9, attrs, v7->number_prototype);
+  v7_set_property(v7, num, "constructor", 11, attrs, ctor);
```
mmikulicic: `constructor` is not a property of itself.

From the point of view of JS, the constructor function is `num`, a plain object with is callable through some magic.

The constructor property should be set in the prototype (I'd group it together with the other `xxxxxx(v7, v7->number_prototype....` functions):

```js
v7_set_property(v7, v7->number_prototype, "constructor", 11, attrs, num);
```

This should fix a bunch of regressions.

src/interpreter.c
```diff
@@ -696,7 +696,7 @@ static val_t i_eval_expr(struct v7 *v7, struct ast *a, ast_off_t *pos,
       }
     case AST_INSTANCEOF:
       v1 = i_eval_expr(v7, a, pos, scope);
-      v2 = i_eval_expr(v7, a, pos, scope);
+      v2 = i_value_of(v7, i_eval_expr(v7, a, pos, scope));
```
mmikulicic: What is the purpose of this?

mmikulicic: Ah, I see. This will return the cfunction embedded in the callable wrapper. But this cfunction doesn't pass `v7_is_function()`.

Moreover, below the attribute `prototype` is read from `v2` which should be the wrapper, not the cfunc which cannot hold properties.

Why not:

```js
      v1 = i_eval_expr(v7, a, pos, scope);
      v2 = i_eval_expr(v7, a, pos, scope);
      if (!v7_is_function(v2) && !v7_is_cfunction(i_value_of(v7, v2))) {
        throw_exception(v7, "TypeError",
                        "Expecting a function in instanceof check");
      }
      return v7_create_boolean(is_prototype_of(v1,
                                               v7_get(v7, v2, "prototype", 9)));
```

src/number.c
cpq: Done.

src/interpreter.c
cpq: Yeah that's better. Changed.

src/number.c
```diff
@@ -80,22 +81,18 @@ static val_t n_isNaN(struct v7 *v7, val_t this_obj, val_t args) {
...
-          "                        {value:__Number_ctor});"
-          "   return this.__ctor(v)}; "
-          "Number");
-  v7->number_prototype = v7_get(v7, num, "prototype", 9);
-  v7_get_property(num, "prototype", 9)->attributes |= V7_PROPERTY_READ_ONLY;
-
-  v7_set_property(v7, v7->global_object, "__Number_ctor", 13, attrs, ctor);
+  v7_set_property(v7, num, "", 0, V7_PROPERTY_HIDDEN, ctor);
+  v7_set_property(v7, num, "prototype", 9, attrs, v7->number_prototype);
+  v7_set_property(v7, v7->number_prototype, "constructor", 11, attrs, ctor);
```
mmikulicic: `s/ctor/num`

This would also fix the annoying `687` ecma test, annoying because the error message contains a pointer and thus it always diffs.

mmikulicic:
After src/number.c:90 we also need a rebase for this to be mergeable, likely the conflict is in the ecma report itself.

cpq:
Done

PR #246, opened 2015-02-11
Add profile generation script

cpq:


tests/generate_profile.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+#
+
+# Make sure you have the tools installed:
+# brew install graphviz gv google-perftools
+
+export CPUPROFILE=/tmp/bin.prof
+make -C tests CFLAGS_EXTRA="-lprofiler -Wl,-no_pie" TEST_FILTER='&& pprof --gv '
```
mmikulicic: why both `pprof --gv` and file generation?

cpq: removed `TEST_FILTER='&& pprof --gv`

PR #258, opened 2015-02-16
15.9 Date Objects - implementation

alashkin:
Date class is implemented in this pull request.
Should close https://github.com/cesanta/v7/issues/132

### DIFF
--- a/src/date.c
+++ b/src/date.c
@@ -23,7 +23,8 @@ typedef double etime_t;
 #define MAX_TIME 8640000000000000
 #define INVALID_TIME NAN
 
-/*** ecma helpers ****/
+/*** ECMA date helpers ****/
+
 #define msPerDay 86400000
 #define HoursPerDay 24
 #define MinutesPerHour 60
@@ -31,6 +32,7 @@ typedef double etime_t;
 #define msPerSecond 1000
 #define msPerMinute 60000
 #define msPerHour 3600000
+#define MonthInYear 12
 
 typedef int64_t etimeint_t;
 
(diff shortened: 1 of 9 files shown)


mmikulicic:
Thanks Alexander!

Would it be a problem for you to rebase this against upstream/master instead of having the merge commit?

alashkin:
I supposed no.. but get this on rebase attempt:
>>error: could not apply 7543333... Date: init, ctor & a couple of functions

I'll try to resolve this.
BTW, why don't you like merge commit?

alashkin:
Close on error =)

mmikulicic:
Our workflow is: feature branch, containing one or more commits, which will then get merged into master with a merge commit.

I think merging back the upstream branch (one or more times) often just introduces unnecessarily complicated history. (practical aesthetics? is there such a thing?)

In your example you probably just wanted to check if everything is all right against master. I think that's the job of the continuous integration.

We usually handle merge conflicts by rebasing instead of merging back the master branch. We don't bother rebasing if there are no merge conflicts.

Of course rebasing is not always feasible, especially if there are many commits to rebase, but I found that:

```
[rerere]
        enabled = 1
```

often helps.

Anyway, don't waste much time on it; on a first look it seemed to me that there were not many conflicts, but a merge commit is not a big deal.

alashkin:
Marco, I've successfully rebased, so, I'm going to close this PR and open another one. Is it OK?

mmikulicic:
Sure, it's ok.

Force update is also an option (it would keep the PR, but github will make the existing comments disappear)

alashkin:
I'll open new PR with rebased sources

PR #260, opened 2015-02-16
15.9 Date Objects - implementation

### LINKED ISSUE #132: Implement Date interface

alashkin:
Date class is implemented in this pull request. Should close #132

Rebased source edition.
Marco's comments from #258 is incorporated.

### DIFF
--- a/src/date.c
+++ b/src/date.c
@@ -1,18 +1,989 @@
 /*
- * Copyright (c) 2014 Cesanta Software Limited
+ * Copyright (c) 2015 Cesanta Software Limited
  * All rights reserved
  */
 
 #include "internal.h"
+#include <sys/time.h>
+#include <time.h>
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <locale.h>
+#include <stddef.h>
 
-V7_PRIVATE enum v7_err Date_ctor(struct v7_c_func_arg *cfa) {
-  struct v7_val *obj = cfa->called_as_constructor ? cfa->this_obj :
-                       v7_push_new_object(cfa->v7);
-  v7_set_class(obj, V7_CLASS_DATE);
-  return V7_OK;
+#ifdef __APPLE__
+int64_t strtoll(const char *, char **, int);
+#else
+extern long timezone;
+extern struct tm *getdate(const char *);
+#endif
+
+typedef double etime_t; /* double is suitable type for ECMA time */
+/* inernally we have to use 64-bit integer for some operations */
+typedef int64_t etimeint_t;
+#define INVALID_TIME NAN
+
+/*++++++  ECMA date & time helpers ++++++*/
+
+#define msPerDay 86400000
+#define HoursPerDay 24
+#define MinutesPerHour 60
+#define SecondsPerMinute 60
+#define SecondsPerHour 3600
+#define msPerSecond 1000
+#define msPerMinute 60000
+#define msPerHour 3600000
+#define MonthsInYear 12
+
+static etimeint_t g_gmtoffms; /* timezone offset, ms, no DST */
+static const char* g_tzname; /* current timezone name */
+
+static etimeint_t ecma_Day(etime_t t) {
+  return floor(t/ msPerDay);
 }
 
-V7_PRIVATE void init_date(void) {
-  init_standard_constructor(V7_CLASS_DATE, Date_ctor);
-  SET_RO_PROP_V(s_global, "Date", s_constructors[V7_CLASS_DATE]);
+/* Leap year formula copied from ECMA 5.1 standart as is */
+static int ecma_DaysInYear(int y) {
+  if (y % 4 != 0 ) {
+    return 365;
+  } else if ( y % 4 == 0 && y % 100 != 0 ) {
+    return 366;
+  } else if (y % 100 == 0 && y % 400 != 0) {
+    return 365;
+  } else if (y % 400 == 0) {
+    return 366;
+  } else {
+    return 365;
+  }
+}
+
+static etimeint_t ecma_DayFromYear(etimeint_t y) {
+  return 365 * (y-1970) + floor((y-1969)/4) - floor((y-1901)/100)
+      + floor((y-1601)/400);
+}
+
+static etimeint_t ecma_TimeFromYear(etimeint_t y) {
+  return msPerDay * ecma_DayFromYear(y);
+}
+
+static int ecma_YearFromTime_s(etime_t t) {
+  int first = floor((t / msPerDay) / 366) + 1970,
+              last = floor((t / msPerDay) / 365) + 1970,
+              middle = 0;
+
+  if (last < first) {
+    int temp = first;
+    first = last;
+    last = temp;
+  }
+
+  while (last > first) {
+    middle = (last + first) / 2;
+    if (ecma_TimeFromYear(middle) > t) {
+      last = middle - 1;
+    } else {
+      if (ecma_TimeFromYear(middle) <= t) {
+        if (ecma_TimeFromYear(middle + 1) > t) {
+          first = middle;
+          break;
+        }
+        first = middle + 1;
+      }
+    }
+  }
+
+  return first;
+}
+
+static int ecma_IsLeapYear(int year) {
+  return ecma_DaysInYear(year) == 366;
+}
+
+static int ecma_DayWithinYear(etime_t t, int year) {
+  return (int)(ecma_Day(t) - ecma_DayFromYear(year));
+}
+
+static int *ecma_getfirstdays(int isleap) {
+  static int sdays[2][MonthsInYear+1] =
+    {{0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334, 365},
+     {0, 31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366}};
+
+  return sdays[isleap];
+}
+
+static int ecma_MonthFromTime(etime_t t, int year) {
+  int *days, i;
+  etimeint_t dwy = ecma_DayWithinYear(t, year);
+
+  days = ecma_getfirstdays(ecma_IsLeapYear(year));
+
+  for (i = 0; i < MonthsInYear; i++) {
+    if (dwy >= days[i] && dwy < days[i+1]) {
+      return i;
+    }
+  }
+
+  return -1;
+}
+
+static int ecma_DateFromTime(etime_t t, int year) {
+  int *days,
+      mft = ecma_MonthFromTime(t, year),
+      dwy = ecma_DayWithinYear(t, year);
+
+  if (mft > 11) {
+    return -1;
+  }
+
+  days = ecma_getfirstdays(ecma_IsLeapYear(year));
+
+  return dwy - days[mft] + 1;
+}
+
+static int ecma_WeekDay(etime_t t) {
+  int ret = (ecma_Day(t)+4) % 7;
+  if (ret < 0) {
+    ret += 7;
+  }
+
+  return ret;
+}
+
+static int ecma_DaylightSavingTA(etime_t t) {
+  time_t time = t / 1000;
+  struct tm tm;
+  memset(&tm, 0, sizeof(t));
+  localtime_r(&time, &tm);
+  if(tm.tm_isdst > 0) {
+    return msPerHour;
+  } else {
+    return 0;
+  }
+}
+
+static int ecma_LocalTZA() {
+  return (int)-g_gmtoffms;
+}
+
+static etimeint_t ecma_LocalTime(etime_t t) {
+  return t + ecma_LocalTZA() + ecma_DaylightSavingTA(t);
+}
+
+static etimeint_t ecma_UTC(etime_t t) {
+  return t - ecma_LocalTZA() - ecma_DaylightSavingTA(t-ecma_LocalTZA());
+}
+
+#define DEF_EXCTRACT_TIMEPART(funcname, c1, c2) \
+  static int ecma_##funcname(etime_t t) { \
+    int ret = (etimeint_t)floor( t / c1) % c2; \
+    if(ret < 0) { \
+      ret += c2; \
+    } \
+    return ret; \
+  }
+
+DEF_EXCTRACT_TIMEPART(HourFromTime, msPerHour, HoursPerDay)
+DEF_EXCTRACT_TIMEPART(MinFromTime, msPerMinute, MinutesPerHour)
+DEF_EXCTRACT_TIMEPART(SecFromTime, msPerSecond, SecondsPerMinute)
+DEF_EXCTRACT_TIMEPART(msFromTime, 1, msPerSecond)
+
+static etimeint_t ecma_MakeTime(etimeint_t hour, etimeint_t min,
+                                etimeint_t sec, etimeint_t ms) {
+  return ((hour * MinutesPerHour + min) * SecondsPerMinute + sec)
+      * msPerSecond + ms;
+}
+
+static etimeint_t ecma_MakeDay(int year, int month, int date) {
+  int *days;
+  etimeint_t yday, mday;
+
+  year += floor(month / 12);
+  month = month % 12;
+  yday = floor(ecma_TimeFromYear(year) / msPerDay);
+  days = ecma_getfirstdays(ecma_IsLeapYear(year));
+  mday = days[month];
+
+  return yday + mday + date - 1;
+}
+
+static etimeint_t ecma_MakeDate(etimeint_t day, etimeint_t time) {
+  return (day * msPerDay + time);
+}
+
+/*++++ ECMA alternatives to struct tm ++++*/
+
+struct timeparts {
+  int year; /* can be negative, up to +-282000 */
+  int month; /* 0-11 */
+  int day; /* 1-31 */
+  int hour; /* 0-23 */
+  int min; /* 0-59 */
+  int sec; /* 0-59 */
+  int msec;
+  int dayofweek; /* 0-6 */
+};
+
+/*+++ this functions is used to get current date/time & timezone */
+
+static void d_gettime(etime_t *time) {
+  struct timeval tv;
+  gettimeofday(&tv, NULL);
+  *time = (etime_t)tv.tv_sec * 1000 + (etime_t)tv.tv_usec / 1000;
+}
+
+static const char *d_gettzname() {
+  return g_tzname;
+}
+
+static etime_t d_mktime_impl(const struct timeparts *tp) {
+  return ecma_MakeDate(ecma_MakeDay(tp->year, tp->month, tp->day),
+                       ecma_MakeTime(tp->hour, tp->min, tp->sec, tp->msec));
+}
+
+/*++++ libc mktime alternatives ++++*/
+
+static etime_t d_lmktime(const struct timeparts *tp) {
+  return ecma_UTC(d_mktime_impl(tp));
+}
+
+static etime_t d_gmktime(const struct timeparts *tp) {
+  return d_mktime_impl(tp);
+}
+
+typedef etime_t (*fmaketime)(const struct timeparts* );
+
+/*++++ libc gmtime & localtime alternatives ++++*/
+static void d_gmtime(const etime_t *t, struct timeparts *tp) {
+  tp->year = ecma_YearFromTime_s(*t);
+  tp->month = ecma_MonthFromTime(*t, tp->year);
+  tp->day = ecma_DateFromTime(*t, tp->year);
+  tp->hour = ecma_HourFromTime(*t);
+  tp->min = ecma_MinFromTime(*t);
+  tp->sec = ecma_SecFromTime(*t);
+  tp->msec = ecma_msFromTime(*t);
+  tp->dayofweek = ecma_WeekDay(*t);
+}
+
+static void d_localtime(const etime_t* time, struct timeparts *tp) {
+  etime_t local_time = ecma_LocalTime(*time);
+  d_gmtime(&local_time, tp);
+}
+
+typedef void (*fbreaktime)(const etime_t* , struct timeparts* );
+
+/*++++ some isXXX helpers ++++*/
+static val_t d_trytogetobjforstring(struct v7 *v7, val_t obj) {
+  val_t ret = i_value_of(v7, obj);
+  if (ret == V7_TAG_NAN) {
+    throw_exception(v7, "TypeError", "Date is invalid (for string)");
+  }
+  return ret;
+}
+
+static int d_iscalledasfunction(struct v7 *v7, val_t this_obj) {
+  /* TODO(alashkin): verify this statement */
+  return is_prototype_of(this_obj, v7->date_prototype);
+}
+
+/*++++ from/to string helpers ++++*/
+
+static int d_timetoISOstr(const etime_t *time, char *buf, size_t buf_size) {
+  /* ISO format: "+XXYYYY-MM-DDTHH:mm:ss.sssZ"; */
+  struct timeparts tp;
+  char use_ext = 0;
+  const char* ey_frm = "%06d-%02d-%02dT%02d:%02d:%02d.%03dZ";
+  const char* simpl_frm = "%d-%02d-%02dT%02d:%02d:%02d.%03dZ";
+
+  d_gmtime(time, &tp);
+
+  if (abs(tp.year) > 9999 || tp.year < 0) {
+    *buf = (tp.year > 0)? '+' : '-';
+    use_ext = 1;
+  }
+
+  return snprintf(buf + use_ext, buf_size - use_ext,
+                  use_ext? ey_frm : simpl_frm, abs(tp.year), tp.month + 1,
+                  tp.day, tp.hour, tp.min, tp.sec, tp.msec) + use_ext;
+}
+
+/* non-locale function should always return in english and 24h-format */
+static const char* wday_name[] = {
+  "Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"
+};
+
+static const char* mon_name[] = {
+  "Jan", "Feb", "Mar", "Apr", "May", "Jun",
+  "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"
+};
+
+int d_getnumbyname(const char **arr, int arr_size, const char *str) {
+  int i;
+  for (i = 0; i < arr_size; i++) {
+    if (strncmp(arr[i], str, 3) == 0 ) {
+      return i + 1;
+    }
+  }
+
+  return -1;
+}
+
+int date_parse(const char *str, int *a1, int *a2, int *a3,
+               char sep, char *rest) {
+  char frmDate[] = " %d/%d/%d%[^\0]";
+  frmDate[3] = frmDate[6] = sep;
+  return sscanf(str, frmDate, a1, a2, a3, rest);
+}
+
+#define NO_TZ 0x7FFFFFFF
+
+/*
+ * not very smart but simple, and working according
+ * to ECMA5.1 StringToDate function
+ */
+
+static int d_parsedatestr(const char *str, struct timeparts *tp, int *tz) {
+  char gmt[4];
+  char buf1[100] = {0}, buf2[100] = {0};
+  int res = 0;
+  memset(tp, 0, sizeof(*tp));
+  *tz = NO_TZ;
+
+  /* #1: trying toISOSrting() format */
+  {
+    const char *frmISOString = " %d-%02d-%02dT%02d:%02d:%02d.%03dZ";
+    res = sscanf(str, frmISOString, &tp->year, &tp->month, &tp->day,
+                 &tp->hour, &tp->min, &tp->sec, &tp->msec);
+    if (res == 7) {
+      *tz = 0;
+      return 1;
+    }
+  }
+
+  /* #2: trying getdate() - it never works on some OS, but... */
+  {
+    struct tm *tm = getdate(str);
+    if (tm != NULL) {
+      tp->year = tm->tm_year + 1900;
+      tp->month = tm->tm_mon;
+      tp->day = tm->tm_mday;
+      tp->hour = tm->tm_hour;
+      tp->min = tm->tm_min;
+      tp->sec = tm->tm_sec;
+
+      return 1;
+    }
+  }
+
+  /* #3: trying toString()/toUTCString()/toDateFormat() formats */
+  {
+    char month[4];
+    const char* frmString = " %03*s %03s %02d %d %02d:%02d:%02d %03s%d";
+    res = sscanf(str, frmString, month, &tp->day, &tp->year,
+                 &tp->hour, &tp->min, &tp->sec, gmt, tz);
+    if (res == 3 || (res >= 6 && res <= 8) ) {
+      if ( (tp->month = d_getnumbyname(mon_name,
+                                       ARRAY_SIZE(mon_name), month)) != -1) {
+        if (res == 7 && strncmp(gmt, "GMT", 3) == 0 ) {
+          *tz = 0;
+        }
+        return 1;
+      }
+    }
+  }
+
+  /* #4: trying the rest */
+
+  /* trying date */
+
+  if (!(date_parse(str, &tp->month, &tp->day, &tp->year, '/', buf1) >= 3 ||
+       date_parse(str, &tp->day, &tp->month, &tp->year, '.', buf1) >= 3 ||
+       date_parse(str, &tp->year, &tp->month, &tp->day, '-', buf1) >= 3 ) ) {
+    return 0;
+  }
+
+  /*  there is date, trying time; from here return 0 only on errors */
+
+  /* trying HH:mm */
+  {
+    const char *frmMMhh = " %d:%d%[^\0]";
+    res = sscanf(buf1, frmMMhh, &tp->hour, &tp->min, buf2);
+    /* can't get time, but have some symbols, assuming error */
+    if (res < 2 ) {
+      return (strlen(buf2) == 0);
+    }
+  }
+
+  /* trying seconds */
+  {
+    const char* frmss = ":%d%[^\0]";
+    memset(buf1, 0, sizeof(buf1));
+    res = sscanf(buf2, frmss, &tp->sec, buf1);
+  }
+
+  /* even if we don't get seconds we gonna try to get tz */
+  {
+    char* rest = res? buf1 : buf2;
+    char* buf = res? buf2 : buf1;
+    const char* frmtz = " %03s%d%[^\0]";
+
+    res = sscanf(rest, frmtz, gmt, tz, buf);
+    if (res == 1 && strncmp(gmt, "GMT", 3) == 0) {
+      *tz = 0;
+    }
+  }
+
+  /* return OK if we are at the end of string */
+  return res <= 2;
+}
+
+static int d_timeFromString(etime_t *time, const char *str, size_t buf_size) {
+  struct timeparts tp;
+  int tz;
+
+  *time = INVALID_TIME;
+
+  if (buf_size > 100) {
+    /* too long for valid date string */
+    return 0;
+  }
+
+  if (d_parsedatestr(str, &tp, &tz)) {
+    /* check results */
+    int valid = 0;
+
+    tp.month--;
+    valid = tp.day >=1 && tp.day <= 31;
+    valid &= tp.month >=0 && tp.month <= 11;
+    valid &= tp.hour >= 0 && tp.hour <= 23;
+    valid &= tp.min >= 0 && tp.min <= 59;
+    valid &= tp.sec >= 0 && tp.sec <= 59;
+
+    if (tz != NO_TZ && tz > 12) {
+      tz /= 100;
+    }
+
+    valid &= (abs(tz) <= 12 || tz == NO_TZ);
+
+    if (valid) {
+      *time = d_gmktime(&tp);
+
+      if (tz != NO_TZ) {
+        /* timezone specified, use it */
+        *time -= tz * msPerHour;
+      } else if (tz != 0 ) {
+        /* assuming local timezone and moving back to UTC */
+        *time = ecma_UTC(*time);
+      }
+    }
+  }
+
+  return !isnan(*time);
+}
+
+typedef int (*ftostring)(const struct timeparts*, char*, int);
+
+static val_t d_tostring(struct v7 *v7, val_t obj, fbreaktime breaktimefunc,
+                        ftostring tostringfunc, int addtz) {
+  struct timeparts tp;
+  int len;
+  char buf[100];
+  etime_t time;
+
+  time = v7_to_double(d_trytogetobjforstring(v7, obj));
+
+  breaktimefunc(&time, &tp);
+  len = tostringfunc(&tp, buf, addtz);
+
+  return v7_create_string(v7, buf, len, 1);
+}
+
+struct d_locale {
+  char locale[50];
+};
+
+static void d_getcurrentlocale(struct d_locale *loc) {
+  strcpy(setlocale(LC_TIME, 0), loc->locale);
+}
+
+static void d_setlocale(const struct d_locale *loc) {
+  setlocale(LC_TIME, loc? loc->locale : "");
+}
+
+/* TODO(alashkin): check portability */
+static val_t d_tolocalestr(struct v7 *v7, val_t obj, const char *frm) {
+  char buf[250];
+  size_t len;
+  struct tm t;
+  etime_t time;
+  struct d_locale prev_locale;
+  struct timeparts tp;
+
+  time = v7_to_double(d_trytogetobjforstring(v7, obj));
+
+  d_getcurrentlocale(&prev_locale);
+  d_setlocale(0);
+  d_localtime(&time, &tp);
+
+  memset(&t, 0, sizeof(t));
+  t.tm_year = tp.year - 1900;
+  t.tm_mon = tp.month;
+  t.tm_mday = tp.day;
+  t.tm_hour = tp.hour;
+  t.tm_min = tp.min;
+  t.tm_sec = tp.sec;
+  t.tm_wday = tp.dayofweek;
+
+  len = strftime(buf, sizeof(buf), frm, &t);
+
+  d_setlocale(&prev_locale);
+
+  return v7_create_string(v7, buf, len, 1);
+}
+
+static int d_tptodatestr(const struct timeparts *tp, char *buf, int addtz) {
+  (void) addtz;
+
+  return sprintf(buf, "%s %s %02d %d", wday_name[tp->dayofweek],
+                 mon_name[tp->month], tp->day, tp->year);
+}
+
+static int d_tptotimestr(const struct timeparts *tp, char *buf, int addtz) {
+  int len;
+
+  len = sprintf(buf, "%02d:%02d:%02d GMT", tp->hour, tp->min, tp->sec);
+
+  if (addtz && g_gmtoffms != 0) {
+    len = sprintf(buf+len, "%c%02d00 (%s)", g_gmtoffms > 0? '-' : '+',
+                  abs((int)g_gmtoffms/msPerHour), d_gettzname());
+  }
+
+  return (int)strlen(buf);
+}
+
+static int d_tptostr(const struct timeparts *tp, char *buf, int addtz) {
+  int len = d_tptodatestr(tp, buf, addtz);
+  *(buf + len) = ' ';
+  return d_tptotimestr(tp, buf + len + 1, addtz) + len + 1;
+}
+
+/*+++ setXXXX & getXXXX helpers ++++*/
+static struct timeparts* d_getTP(val_t val,  struct timeparts *tp,
+                                 fbreaktime breaktimefunc) {
+  etime_t time;
+  time = v7_to_double(val);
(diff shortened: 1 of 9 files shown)


alashkin:
I had problems during rebasing. Hope nothing lost (and nothing "new" found).

mmikulicic:
Rewriting history comes with great power but great responsibility :-)

I'll review it tomorrow. Perhaps it's better to do the review on the whole
PR diff since you fixed many details in later commits.

alashkin:
Yes, I changed things like «Range Error» then I started to pass ecma-tests.
So, only the latest commit (of date.c) is correct version.

> 16 февр. 2015 г., в 19:42, Marko Mikulicic <notifications@github.com> написал(а):
> 
> Rewriting history comes with great power but great responsibility :-)
> 
> I'll review it tomorrow. Perhaps it's better to do the review on the whole
> PR diff since you fixed many details in later commits.

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+
+typedef void (*fbreaktime)(etime_t* , struct timeparts* );
+
+/*++++ some isXXX helpers ++++*/
+
+static int d_isnumberNAN(struct v7 *v7, val_t obj) {
+  return (i_value_of(v7, obj) == V7_TAG_NAN);
+}
+
+static void d_isobjvalidforstirng(struct v7* v7, val_t obj) {
```
mmikulicic: typo in function name.

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+  return is_prototype_of(this_obj, v7->date_prototype);
+}
+
+/*++++ from/to string helpers ++++*/
+
+static int d_timetoISOstr(etime_t* time, char* buf, size_t buf_size) {
+  /* ISO format: "+XXYYYY-MM-DDTHH:mm:ss.sssZ"; */
+  struct timeparts tp;
+  char use_ext = 0;
+  const char* ey_frm="%06d-%02d-%02dT%02d:%02d:%02d.%03dZ";
```
mmikulicic: Style nitpick: Throughout our code base we use `char *v` instead of `char* v`.
The latter might seem more natural if you think about `<type> <varname>` but C variable declaration syntax is more fun than that.

src/date.c
```diff
@@ -1,18 +1,1006 @@
 /*
```
mmikulicic: Our codebase tries to follow http://google-styleguide.googlecode.com/svn/trunk/cppguide.html
and for the optional parts we just favour consistency.

I'll focus on other stuff and we'll fix the style later.

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+DEF_TOLOCALESTR(LocaleString, "%c")
+DEF_TOLOCALESTR(LocaleDateString, "%x")
+DEF_TOLOCALESTR(LocaleTimeString, "%X")
+
+/*++++ get functions ++++*/
+
+#define DEF_GET_TP_FUNC(funcName, tpmember, breaktimefunc) \
+  static val_t Date_get##funcName(struct v7 *v7, val_t this_obj, val_t args) { \
+    struct timeparts tp; \
+    return v7_create_number(d_isnumberNAN(v7, this_obj)? NAN: d_getTP(v7, \
```
mmikulicic: why not invoke `i_value_of` only once here?

Rationale:

`i_value_of` takes a value and returns itself if it's a scalar, the result of the method `valueOf()` if it's an object (e.g. a custom method or the builtins such as Number.valueOf() which extracts a scalar from the object wrappers).

In the case of `Date`, calling `i_value_of` on `this_obj` in each method is just a way to extract it's hidden property, like we do for `Number`, `Boolean` and friends.

There is no need to push down this detail into helpers such as `d_isnumberNAN`, especially since by not doing so we'll also remove unnecessary method invocations.

src/date.c
mmikulicic: I'd prefer `check...` instead of `is...` for functions that throw exceptions instead of returning bool.

mmikulicic: `v7_is_object(obj)` will always be true here, and `i_value_of` is the identity function for non-objects, so this is basically

```js
if (i_value_of(v7, obj) == V7_TAG_NAN) {
  throw_exception(v7, "TypeError", "Date is invalid (for string)")
}
```

Why not pass the Date's value (as discussed in another comment) here and just check whether it's NaN ?
Or why not even get rid of this helper altogether since it will be a two-liner with a naming disorder?

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+/*++++ from/to string helpers ++++*/
+
+static int d_timetoISOstr(etime_t* time, char* buf, size_t buf_size) {
+  /* ISO format: "+XXYYYY-MM-DDTHH:mm:ss.sssZ"; */
+  struct timeparts tp;
+  char use_ext = 0;
+  const char* ey_frm="%06d-%02d-%02dT%02d:%02d:%02d.%03dZ";
+  const char* simpl_frm="%04d-%02d-%02dT%02d:%02d:%02d.%03dZ";
+  
+  (void)buf_size;
```
mmikulicic: you do use `buf_size` now, this can be removed

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+  
+  d_gmtime(time, &tp);
+
+  if(abs(tp.year) > 9999 || tp.year < 0) {
+    *buf = tp.year>0? '+':'-';
```
mmikulicic: Doesn't this output a double minus if the year is negative?

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+static etimeint_t ecma_LocalTime(etime_t t) {
+  return t + ecma_LocalTZA() + ecma_DaylightSavingTA(t);
+}
+
+static etimeint_t ecma_UTC(etime_t t) {
+  return t - ecma_LocalTZA() - ecma_DaylightSavingTA(t-ecma_LocalTZA());
+}
+
+static int ecma_HourFromTime(etime_t t) {
+  return (etimeint_t)floor(t / msPerHour) % HoursPerDay;
```
mmikulicic: How does this work with dates before epoch?

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+}
+
+static int ecma_MinFromTime(etime_t t) {
+  return (etimeint_t)floor(t / msPerMinute) % MinutesPerHour;
```
mmikulicic: Same here

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+}
+
+static int ecma_SecFromTime(etime_t t) {
+  return (etimeint_t)floor(t / msPerSecond) % SecondsPerMinute;
```
mmikulicic: And here

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+}
+
+static int ecma_msFromTime(etime_t t) {
+  return (etimeint_t)t % msPerSecond;
```
mmikulicic: Dito

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+/*+++ this functions is used to get current date/time & timezone */
+
+static void d_gettime(etime_t* time) {
+  struct timeval tv;
+  gettimeofday(&tv, NULL);
+  *time = (etime_t)tv.tv_sec * 1000 + (etime_t)tv.tv_usec / 1000;
+}
+
+static int d_gettimezone() {
+  return (int)timezone/60; /* TODO(alashkin): check non-OSX */
```
mmikulicic: Our dev machines are OSX as well. We found a simple way of testing the code on linux:

Install boot2docker and run `make docker`. It will build and run all unit tests from right from your checkout.

mmikulicic: I don't know `timezone` but it sounds weird to me that's defined in minutes rather than seconds. `tm_gmtoff` is defined as seconds.

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+  ecma_getfirstdays(days, ily);
+  
+  return dwy - days[mft] + 1;
+}
+
+static int ecma_WeekDay(etime_t t) {
+  return (ecma_Day(t)+4) % 7;
+}
+
+static int ecma_DaylightSavingTA(etime_t t) {
```
mmikulicic: What does `TA` stand for ?

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+/*++++++  ECMA date & time helpers ++++++*/
+
+#define msPerDay 86400000
+#define HoursPerDay 24
+#define MinutesPerHour 60
+#define SecondsPerMinute 60
+#define msPerSecond 1000
+#define msPerMinute 60000
+#define msPerHour 3600000
+#define MonthInYear 12
```
mmikulicic: I guess you meant `MonthsInYear` here.

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+    for(i = 2; i < ARRAY_SIZE(sdays); i++) {
+      days[i] += 1;
+    }
+  }
+}
+
+static int ecma_MonthFromTime(etime_t t, int year) {
+  int days[MonthInYear+1];
+  etimeint_t dwy = ecma_DayWithinYear(t, year);
+  int ily = ecma_InLeapYear(t, year);
```
mmikulicic: I think calling `ily` something like `leap` is easier to read;
or even just get rid of the temporary: `ecma_getfirstdays(days, ecma_InLeapYear(t, year));`

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+  
+  unsigned int i; int ret = -1;
+  
+  ecma_getfirstdays(days, ily);
```
mmikulicic: Why not just having a two-dimensional static array and index by `ily` ?
The simpler, the better.

src/date.c
alashkin: Renamed to d_checkobjvalidforstring

src/date.c
mmikulicic: I understand that the JS
but I see that internally the code would be cleaner if you return seconds here
(things like `tz = (tz == NO_TZ)? -d_gettimezone() * msPerMinute : tz * msPerHour;` would be cleaner) and transform to minutes when handling `getTimezoneOffset`.

src/date.c
alashkin: Fixed

src/date.c
alashkin: TA = Time Adjustment
ECMA 5.1 15.9.1.8

src/date.c
alashkin: As for timezone:
Date::getTimezoneOffset should return in minutes
For internal operation I need it in ms.
In this matter there is not difference between timezone and tm_gmtoff

src/date.c
alashkin: Fixed

src/date.c
alashkin: Ok

src/date.c
alashkin: I'm not sure that I'm understand your point - take a look on new version, pls

src/date.c
alashkin: d_isnumberNAN was removed
This function was important because I tried to use int64 for time and as result d_isnumberNAN was something like (mytime != MY_MAGIC_NUMBER)
Since I came to double it became obsolete.

src/date.c
alashkin: Why shouldn't?
I use timezone var to detect current timezone only.
The rest is independent of 1970.

src/date.c
alashkin: See previous comment

src/date.c
alashkin: See previous comment

src/date.c
alashkin: ?

src/date.c
alashkin: Temp vars help with debug, sometimes.
But - as you wish. I dropped all ily vars.

src/date.c
```diff
@@ -1,18 +1,1006 @@
...
+static int ecma_LocalTZA() {
+  return (int)-timezone * 1000;
+}
+
+static etimeint_t ecma_LocalTime(etime_t t) {
+  return t + ecma_LocalTZA() + ecma_DaylightSavingTA(t);
+}
+
+static etimeint_t ecma_UTC(etime_t t) {
+  return t - ecma_LocalTZA() - ecma_DaylightSavingTA(t-ecma_LocalTZA());
```
mmikulicic: I wonder how this works around the DST change

src/date.c
mmikulicic: Thanks, makes sense.

src/date.c
mmikulicic: Yes they do; a more explicit (less acronymous) name was an alternative suggestion if you prefer being able to keep this debug point.

src/date.c
alashkin: The simpler, the better, the same...
Fixed in any case.

src/date.c
mmikulicic: Yep, sounds good

src/date.c
mmikulicic: I meant "ditto", and I guess you'd answer "See previous comment" :-)

src/date.c
mmikulicic: Aren't etime_t negative for timestamps before epoch?
In that case the `floor(t / msPerHour) % HoursPerDay` will be a negative number, which is not within the range of legal values you documented in

```c
struct timeparts {
....
  int hour; /* 0-23 */`
```

src/date.c
alashkin: d_gettimezone was removed

src/date.c
mmikulicic: I don't see the new version. How did you push it ?

Anyway, I meant that `d_isnumberNAN` invokes `i_value_of` and also `d_getTP` invokes `i_value_of`. IMHO both functions should work on the number value contained in `this_obj`, instead of taking directly `this_obj`.

A side effect is that i_value_of would get invoked half of the times (which is good since it causes a method invocation, allocation of an JS stack frame etc).

src/date.c
alashkin: OMG!
Wrong copy & paste from ecma document.
Fixed & added to unit tests

src/date.c
alashkin: Should I read it before making git push?

src/date.c
alashkin: I didn't pushed yet.
I'm going to push all changes when you'll say that you complete with first iteration.

src/date.c
mmikulicic: Optional, we can first focus on functional aspects. 
Once we improved the quality of the code by working together on this, and we can
fix the style later.

One thing though: I noticed that your files have trailing whitespaces. Removing whitespaces is a good practice. `git diff` even marks them red.  But the most important thing is that if somebody (e.g. I) have an editor on save hook that removes them on save, it will basically introduce a diff in a dozen places of the `date.c` file on the first save, which is annoying.

src/date.c
mmikulicic: I think the first iteration is complete.

I wish we had a real code review tool instead of this github crap...
For now we managed to survive, because we managed to contain the size of the commits and the number of comments was low, but I'd really like to use something like gerrit.

src/date.c
alashkin: I was confused a bit since according to 
http://www.urbandictionary.com/define.php?term=Dito

Dito
Used commonly by Puerto Ricans or Dominicans as a word to express ptty..like saying "aww poor you"

=))

src/date.c
alashkin: It is very strange since my editor _should_ does the same thing. 
I'll check.

src/date.c
alashkin: I've tried boot2docker, but can't make docker says that tm_gtmoff (and timezone) is undeclared.
I will try again later.

alashkin:
Marco, I've finished with first part of fixes.

Please, PTAL

src/date.c
mmikulicic: lol, I guess most our letter words will have some kind of definition in urbandictionary :-)

v7.c
```diff
@@ -10261,3 +10273,1012 @@ int main(int argc, char *argv[]) {
...
+static etimeint_t ecma_LocalTime(etime_t t) {
+  return t + ecma_LocalTZA() + ecma_DaylightSavingTA(t);
+}
+
+static etimeint_t ecma_UTC(etime_t t) {
+  return t - ecma_LocalTZA() - ecma_DaylightSavingTA(t-ecma_LocalTZA());
+}
+
+static int ecma_HourFromTime(etime_t t) {
+  return abs((etimeint_t)floor(t / msPerHour) % HoursPerDay);
```
mmikulicic: Better, but not right enough :-) 
A screendump is better than a thousand words:

```
$ ./v7 -e 'd=new Date("1969-12-31T23:59:59.999Z"); d.toISOString()'
"1969-11-31T01:01:01.001Z"
```

src/date.c
```diff
@@ -1,18 +1,1012 @@
...
+  ret_time = d_time_number_from_tp(v7, this_obj, args, tpyear, 0, d_gmktime);
+  return v7_create_number(ret_time);
+}
+
+/****** Initialization *******/
+
+/* we should set V7_PROPERTY_DONT_ENUM for all Date props */
+/* TODO(???): check other objects */
+
+static int d_set_cfunc_prop(struct v7 *v7, val_t o, const char *name,
```
mmikulicic: Would `set_cfunc_prop` work as well ?

src/date.c
```diff
@@ -1,18 +1,1012 @@
...
+  }
+
+  ret_time = d_time_number_from_tp(v7, this_obj, args, tpyear, 0, d_gmktime);
+  return v7_create_number(ret_time);
+}
+
+/****** Initialization *******/
+
+/* we should set V7_PROPERTY_DONT_ENUM for all Date props */
+/* TODO(???): check other objects */
```
mmikulicic: Probably it's counterintuitive, but we tend to put always the author of the comment in the `TODO()` mark. It's not the person who has to do it, but rather the person who figured out that there is something to do. The idea is that that person is likely to know more about what and why has to be done.

src/date.c
```diff
@@ -1,18 +1,1012 @@
...
+  etime_t ret_time = 0;   
+  if(v7_is_object(this_obj) && this_obj != v7->global_object) {
+    long cargs = v7_array_length(v7, args);
+    if(cargs <=0 ) {
+      /* no parameters - return current date & time */
+      d_gettime(&ret_time);
+    } else if(cargs == 1) {
+      /* one parameter */
+      val_t arg = v7_array_at(v7, args, 0);
+      if(v7_is_string(arg)){ /* it could be string */
```
mmikulicic: What if it's a string but doesn't parse as a time. Shouldn't it be converted to int?

src/date.c
```diff
@@ -1,18 +1,1012 @@
...
+  }
+  
+  return *ret != INVALID_TIME;
+}
+
+/*+++ setXXXX & getXXXX helpers ++++*/
+static struct timeparts* d_getTP(val_t val, val_t args,
+                            struct timeparts *tp, fbreaktime breaktimefunc) {
+  etime_t time;
+  (void)args;
```
mmikulicic: if `args` are not needed, let's remove them from the function signature.

src/date.c
```diff
@@ -1,18 +1,1012 @@
...
+  *(buf + len) = ' ';
+  return d_tptotimestr(tp, buf + len + 1, addtz) + len + 1;
+}
+
+/* +++ converting arg to string ++++*/
+/* according ECMA 5.1 we should try to convert any type of 
+ * parameter to number before goint NaN 
+*/
+
+static int d_argtoint(struct v7 *v7, val_t *obj, etime_t *ret) {
```
mmikulicic: I think `i_as_num` should do exactly this

src/date.c
```diff
@@ -1,18 +1,1012 @@
...
+static void d_localtime(etime_t* time, struct timeparts *tp) {
+  etime_t local_time = ecma_LocalTime(*time);
+  d_gmtime(&local_time, tp);
+}
+
+typedef void (*fbreaktime)(etime_t* , struct timeparts* );
+
+/*++++ some isXXX helpers ++++*/
+/*
+static int d_isnumberNAN(struct v7 *v7, val_t obj) {
```
mmikulicic: Let's remove dead code

mmikulicic:
General comment about `i_value_of`.

Take a look at how we do it in `Str_split` for example.
We first extract the value with `i_value_of` out of `this` and then pass that value around to all helpers. It makes it easier to understand which kind of values are being passed around and as a side effect it reduces the number of times `i_value_of` (expensive) is invoked.

src/date.c
```diff
@@ -1,18 +1,1012 @@
...
+static val_t d_tostring(struct v7 *v7, val_t obj, fbreaktime breaktimefunc,
+                                        ftostring tostringfunc, int addtz) {
+  struct timeparts tp;
+  int len;
+  char buf[100];
+  etime_t time;
+  
+  d_checkobjvalidforstring(v7, obj);
+  
+  time = i_as_num(v7, obj);
```
mmikulicic: Once `d_tostring` will receive the `value` instead of `this_obj`, this can become a `v7_to_double`.

mmikulicic:
I'm sorry to bother you with this very long review. Things like that happen what the code to be reviewed is very long and when it's the first time the reviewer and reviewee work together.

I'll stop here at this round. Let's address the i_value_of and the pre-epoch bug and when they are addressed we can move on cleaning up code style and I think we're done.

The contribution otherwise looks good to me and if other things pop out we'll can solve them as we go.

src/date.c
alashkin: I've buried its.

src/date.c
alashkin: Great func! 
Yes, exactly what I needed.

src/date.c
alashkin: Done.

src/date.c
alashkin: I tried. But I got tons of messages like "properties XXXX have DontEnum attribute" when running ecma-tests.
That's why I wrote my own func and add TODO(???).
I really don't know who has to resolve this TODO.

src/date.c
alashkin: See previous comment. I can change "???" to "mkm", if you wish =)

src/date.c
alashkin: It should :(
Done.

src/date.c
alashkin: I got your idea, and walked through the code and fix functions with non-single i_value_of call. 
But I kept the very this place (d_tostring) as is, because it is called directly from Date API functions (ie macroses) and there isn't much sense to change it.

alashkin:
I'm trying again.

PTAL, please.

src/date.c
```diff
@@ -1,18 +1,983 @@
...
+      /* one parameter */
+      val_t arg = v7_array_at(v7, args, 0);
+      if(v7_is_string(arg)){ /* it could be string */
+        size_t str_size;
+        const char* str = v7_to_string(v7, &arg, &str_size);
+        if(!d_timeFromString(&ret_time, str, str_size)) {
+          /* is this string is not valid date string,
+           trying to convert it to number */
+          char* endptr;
+          ret_time = strtoll(str, &endptr, 10);
```
mmikulicic: Could we use `i_as_num` here as well?

I noticed that v8 doesn't accept `new Date("1.0")` while if we use i_as_num here it would convert it to `1`; but I don't know if the ecma standard regulates that.

src/date.c
mmikulicic: Sorry I missed that comment.

src/date.c
mmikulicic: What I meant is that you should put your own nick even if you're necessarily going to be the one who's doing the task (that's why it's counterintuitive).
However a reviewer should also know the context, so it's fine for me if you put mkm :-)

src/date.c
alashkin: Rhino doesn't accept as well.
I think i_as_num will work here, since standard says to convert to number if cannot parse as date- string.
Done.

src/date.c
mmikulicic: btw, cpq made some changes to set_cfunc_prop upstream. We can see after realigning to master.

src/date.c
```diff
@@ -1,18 +1,983 @@
...
+  int *days;
+  etimeint_t dwy = ecma_DayWithinYear(t, year);
+  
+  unsigned int i; int ret = -1;
+  
+  days = ecma_getfirstdays(ecma_InLeapYear(t, year));
+  
+  for(i = 0; i < MonthsInYear; i++) {
+    if(dwy >= days[i] && dwy < days[i+1]) {
+      ret = i;
```
mmikulicic: minor: just `return i` here and `return -1` at the end.

src/date.c
alashkin: I've put  "mkm" =)

src/date.c
```diff
@@ -1,18 +1,983 @@
...
+  etimeint_t yday;
+  etimeint_t mday;
+  
+  year += floor(month / 12);
+  
+  month = month % 12;
+  
+  yday = floor(ecma_TimeFromYear(year) / msPerDay);
+
+  days = ecma_getfirstdays(ecma_DaysInYear(year) == 366);
```
mmikulicic: consistency: either use inLeapYear everywhere or inline this expression (I have a slight preference on the latter)

src/date.c
alashkin: May be - I asked him about this behavior...

src/date.c
```diff
@@ -1,18 +1,983 @@
...
+        }
+        first = middle + 1;
+      }
+    }
+  }
+  
+  return first;
+}
+
+static int ecma_InLeapYear(etime_t t, int year) {
```
mmikulicic: pls remove the `t` argument (or the whole ecma_InLeapYear function, see other comment)

src/date.c
alashkin: Are you sure?
I don't like many returns from function.
Ready to fix if that is code-rule..

mmikulicic:
Amalgamation is a PITA. It seems that you didn't commit the regenerated v7.c file.

Unfortunately git hooks cannot be enforced server side on github.
You can setup your local hooks with:

```make setup-hooks```

File: src/date.c (hunk 28 above)
Comment (alashkin) [inline src/date.c]: t removed

File: src/date.c (hunk 27 above)
Comment (alashkin) [inline src/date.c]: I'd like to keep function instead of inlining.
I can convert it to macro, if you want.

**Comment by alashkin** [comment]
I deleted v7.c from commit.
Why you need it? 
Looks like it gives nothing but conflicts....?

Put updated v7.c to my fork?

File: src/date.c (hunk 27 above)
Comment (mmikulicic) [inline src/date.c]: Function is fine, it's autodocumenting. Consistency is more important.

Comment (alashkin) [inline src/date.c]: Done with function)

**Comment by mmikulicic** [comment]
w.r.t v7.c ... we didn't find a good solution for it yet.

It's a generated file. We want people to be able to embed v7 by just including two files in their projects. We can decide to not keep the generated file in the repo and just distribute it in releases, but for know we kinda like the simplicity of being able to just pull the files straight from the source git/tarball whatever.

Conflicts are annoying but not hard to deal with, since you just have to regenerate the file and git add. More could be done with git hooks, but on the other hand managing hooks can be a pain in the ass.

**Comment by alashkin** [comment]
Yes, I'm understand about v7.c

I just think that it's easier to merge without it, that regenerate v7.c and re-commit it.
But of course, it it not a problem to push my generated version.

Push it right now?

File: src/date.c (hunk 26 above)
Comment (mmikulicic) [inline src/date.c]: Both approaches have their advantages.
We don't have specific fine grained rules about it; the general idea is to not create complicated control flow just to abide to some kind of abstract structured programming practices (unless they buy you something tangible).

In this case, given the nesting level, I find the extra brain cycles needed to check whether this break actually breaks the loop or if there is another loop to not be worth. And the less variables the better.

In other places we even use gotos to jump to common cleanup code.

Comment (alashkin) [inline src/date.c]: Ok, done

**Comment by mmikulicic** [comment]
Yes pushing now is ok.

Any suggestions for how to handle the amalgamated file is welcome :-)

**Comment by alashkin** [comment]
v7 & some fixes pushed.

PTAL again, pls.

**Comment by alashkin** [comment]
>> Failed — AppVeyor was unable to build non-mergeable pull request

Hmmm.. what does that mean?

File: src/date.c [hunk 29]
```diff
@@ -1,18 +1,984 @@
...
+  d_gmtime(&local_time, tp);
+}
+
+typedef void (*fbreaktime)(const etime_t* , struct timeparts* );
+
+/*++++ some isXXX helpers ++++*/
+static val_t d_trytogetobjforstring(struct v7 *v7, val_t obj) {
+  val_t ret = V7_TAG_NAN;
+
+  if(!v7_is_object(obj) || (v7_is_object(obj)
```
mmikulicic:
```c
static val_t d_trytogetobjforstring(struct v7 *v7, val_t this_obj) {
  val_t ret = i_value_of(v7, obj);
  if (ret == V7_TAG_NAN) {
    throw_exception(v7, "TypeError", "Date is invalid (for string)");
  }
  return ret;
}
```

Rationale: 
 a) document that the input value is the Date object.
 b) simplify since it's always the Date object.

mmikulicic:
Appveyor is a CI tool for windows, that doesn't work very well. I've disabled it for non master builds but it still complains when a branch is not mergeable (because of conflicts) against master.

mmikulicic:
End of iteration, no more comments. Please address the outstanding commen
You can also merge back upstream master and fix conflicts.

Then I'd rather squash the whole branch into one single commit, force push and we can iterate on making the style consistent with the rest of V7.

src/date.c
alashkin: Ok, I just wasn't sure what I can call i_value_of(...)  if  !v7_is_object(obj)

mmikulicic:
BTW, you should also commit the diffs in tests/ecma_report.txt

src/date.c
mmikulicic: `i_value_of` works also for non objects, it that case is just the identity function.

mmikulicic: The naming might be confusing. It basically means "hey interpreter, invoke valueOf() on this value". (more or less, since it handles null and undefined as well)

alashkin:
Marco,
Rebased against upstream/master, merged into 1 commit, ecma_report.txt added

PTAL?

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+  return sprintf(buf, "%s %s %02d %d", wday_name[tp->dayofweek],
+                                    mon_name[tp->month], tp->day, tp->year);
+}
+
+static int d_tptotimestr(const struct timeparts *tp, char *buf, int addtz) {
+  int len;
+
+  len = sprintf(buf, "%02d:%02d:%02d GMT", tp->hour, tp->min, tp->sec);
+  
+  if(addtz && g_gmtoffms != 0) {
```
mmikulicic: Space between `if` and `(`

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+
+  len = strftime(buf, sizeof(buf), frm, &t);
+  
+  d_setlocale(&prev_locale);
+  
+  return v7_create_string(v7, buf, len, 1);
+}
+
+static int d_tptodatestr(const struct timeparts *tp, char *buf, int addtz) {
+  (void)addtz;
```
mmikulicic: `(void) addtz`

cpq:
This review is so far an absolute leader by the number of rounds & comments.
Well done gents

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+}
+
+static int d_tptotimestr(const struct timeparts *tp, char *buf, int addtz) {
+  int len;
+
+  len = sprintf(buf, "%02d:%02d:%02d GMT", tp->hour, tp->min, tp->sec);
+  
+  if(addtz && g_gmtoffms != 0) {
+    len = sprintf(buf+len, "%c%02d00 (%s)", g_gmtoffms > 0? '-':'+',
+                                abs((int)g_gmtoffms/msPerHour), d_gettzname());
```
mmikulicic: align at function open paren:

```c
    len = sprintf(buf+len, "%c%02d00 (%s)", g_gmtoffms > 0 ? '-' : '+',
                  abs((int) g_gmtoffms/msPerHour), d_gettzname());
```

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+  breaktimefunc(&time, tp);
+  return tp;
+}
+
+/* notice: holding month in calendar format (1-12, not 0-11) */
+struct dtimepartsarr {
+  etime_t args[7];
+};
+
+enum detimepartsarr { tpyear = 0, tpmonth , tpdate , tphours, tpminutes,
```
mmikulicic: Align or:
```c
enum detimepartsarr {
  tpyear = 0, tpmonth , tpdate , tphours, tpminutes, tpseconds, tpmsec, tpmax
};
```

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+
+static etime_t d_changepartoftime(const etime_t *current,
+                            struct dtimepartsarr *a, fbreaktime breaktimefunc,
+                                            fmaketime maketimefunc) {
+  /* 0 = year, 1 = month , 2 = date , 3 = hours, 
+                4 = minutes, 5 = seconds, 6 = ms */
+  struct timeparts tp;
+  unsigned long i;
+  
+  int* tp_arr[7]; /* C89 doesn't handle initialization
```
mmikulicic: Multiline comments before statement:

```c
/*
 * blah
 * blah
 */
`

File: src/date.c [hunk 35]
```diff
@@ -1,18 +1,981 @@
...
+            int start_pos, fbreaktime breaktimefunc, fmaketime makefilefunc) {
+  etime_t ret_time = INVALID_TIME;
+  long cargs;
+
+  val_t objtime = i_value_of(v7, this_obj);
+
+  if((cargs=v7_array_length(v7, args)) >= 1 && objtime != V7_TAG_NAN) {
+    int i;
+    struct dtimepartsarr a = {{INVALID_TIME, INVALID_TIME, INVALID_TIME,
+                    INVALID_TIME, INVALID_TIME, INVALID_TIME, INVALID_TIME}};
```
mmikulicic: Align

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+        ret_time = i_as_num(v7, arg);
+      }
+    } else {
+      /* 2+ paramaters - should be parts of a date */
+      struct dtimepartsarr a;
+      int i;
+      
+      memset(&a, 0, sizeof(a));
+      
+      for(i=0; i < cargs; i++) {
```
mmikulicic: Spaces around `=` as well

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+        a.args[i] = i_as_num(v7, v7_array_at(v7, args, i));
+        if(isnan(a.args[i])) {
+          break;
+        }
+      }
+      
+      if(i>=cargs) {
```
mmikulicic:
```c
if (i >= cargs) {
```

mmikulicic:
@alashkin the code looks good to me, let's just do a round of restyling and we can close this off :-)

alashkin:
@cpq, yes....
Is it my last PR? =)

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+    } \
+    return ret; \
+  }
+
+DEF_EXCTRACT_TIMEPART(HourFromTime, msPerHour, HoursPerDay)
+DEF_EXCTRACT_TIMEPART(MinFromTime, msPerMinute, MinutesPerHour)
+DEF_EXCTRACT_TIMEPART(SecFromTime, msPerSecond, SecondsPerMinute)
+DEF_EXCTRACT_TIMEPART(msFromTime, 1, msPerSecond)
+
+
```
mmikulicic: remove unnecessary space.

Rationale: visual formatting in the code tends to be subjective and make it easy to introduce non-semantic diffs when people edit the code. (This is one of the reasons consistent style is important in general)

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+static etimeint_t ecma_MakeDay(int year, int month, int date)
+{
+  int *days;
+
+  etimeint_t yday;
+  etimeint_t mday;
+  
+  year += floor(month / 12);
+
```
mmikulicic: pls remove these empty lines as well, too much emphasis; it yells:
```
be

very

careful!

very

weird

things

happen

here
```

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+  /* trying seconds */
+  {
+    const char* frmss = ":%d%[^\0]";
+    memset(buf1, 0, sizeof(buf1));
+    res = sscanf(buf2, frmss, &tp->sec, buf1);
+  }
+  
+  /* even if we don't get seconds we gonna try to get tz */
+  {
+    char* rest = res? buf1: buf2;
```
mmikulicic: spaces around both sides (unlike commas):

`res?  buf1 : buf2;`

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+       date_parse(str, &tp->day, &tp->month, &tp->year, '.', buf1) >= 3 ||
+       date_parse(str, &tp->year, &tp->month, &tp->day, '-', buf1) >= 3 ) ) {
+    return 0;
+  }
+  
+  /*  there is date, trying time; from here return 0 only on errors */
+  
+  /* trying HH:mm */
+  {
+    const char* frmMMhh = " %d:%d%[^\0]";
```
mmikulicic: Pointer goes with variable:

`const char *frmMMhh`

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+    const char* frmtz = " %03s%d%[^\0]";
+    
+    res = sscanf(rest, frmtz, gmt, tz, buf);
+    if(res == 1 && strncmp(gmt, "GMT", 3) == 0) {
+      *tz = 0;
+    }
+  }
+
+  /* return OK if we are at the end of string */
+  return (res <= 2);
```
mmikulicic: pls remove parenthesis

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+    len = sprintf(buf+len, "%c%02d00 (%s)", g_gmtoffms > 0? '-':'+',
+                                abs((int)g_gmtoffms/msPerHour), d_gettzname());
+  }
+  
+  return (int)strlen(buf);
+}
+
+static int d_tptostr(const struct timeparts *tp, char *buf, int addtz) {
+  int len = d_tptodatestr(tp, buf, addtz);
+  *(buf + len) = ' ';
```
mmikulicic: why not just
```c
buf[len] = ' ';
```
?

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+                                breaktimefunc, makefilefunc);
+    }
+  }
+  
+  return ret_time;
+}
+
+
+static val_t d_setTimePart(struct v7 *v7, val_t this_obj, val_t args,
+            int start_pos, fbreaktime breaktimefunc, fmaketime makefilefunc) {
```
mmikulicic: align and possibly split to a third line if needed (looks like it will)

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+      
+      if(i>=cargs) {
+        if(a.args[tpdate] == 0 ) {
+          a.args[tpdate] = 1; /* If date is supplied then let 
+                               dt be ToNumber(date); else let dt be 1. */
+        }
+      
+        if(a.args[tpyear] >=0 && a.args[tpyear] <=99) {
+          a.args[tpyear] += 1900;
+          /* If y is not NaN and 0 <= ToInteger(y) <= 99,
```
mmikulicic: BSD style multiline comments everywhere

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+  static val_t Date_to##funcname(struct v7 *v7, val_t this_obj, val_t args) {  \
+    (void)args; \
+    return d_tostring(v7, this_obj, breaktimefunc, tostrfunc, addtz); \
+  }
+
+DEF_TOSTR(UTCString, d_gmtime, d_tptostr, 0)
+DEF_TOSTR(String, d_localtime, d_tptostr, 1)
+DEF_TOSTR(DateString, d_localtime, d_tptodatestr, 1)
+DEF_TOSTR(TimeString, d_localtime, d_tptotimestr, 1)
+
```
mmikulicic: remove line here as well pls

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+    val_t v = i_value_of(v7, this_obj); \
+    (void)args; \
+    return v7_create_number(v == V7_TAG_NAN? NAN: d_getTP( \
+                                  v, &tp, breaktimefunc)->tpmember); \
+  }
+
+#define DEF_GET_TP(funcName, tpmember) \
+  DEF_GET_TP_FUNC(UTC##funcName, tpmember, d_gmtime) \
+  DEF_GET_TP_FUNC(funcName, tpmember, d_localtime)
+
```
mmikulicic: remove line here as well pls

src/date.c
```diff
@@ -1,18 +1,981 @@
...
+  val_t n;
+  if(v7_array_length(v7, args) >= 1) {
+    ret_time = i_as_num(v7, v7_array_at(v7, args, 0));
+  }
+
+  n = v7_create_number(ret_time);
+  v7_set_property(v7, this_obj, "", 0, V7_PROPERTY_HIDDEN, n);
+  return n;
+}
+
```
mmikulicic: remove line here as well pls

mmikulicic:
Ok, I threw a bunch of style comments, you get the idea.

Awesome! We have dates and 35.something % of ecma coverage!

alashkin:
How should I align things like 
static etimeint_t ecma_DayFromYear(etimeint_t y) {
  return 365 * (y-1970) + floor((y-1969)/4) - floor((y-1901)/100)
                                              + floor((y-1601)/400);
}

?

mmikulicic:
Yep this things are fuzzy. I have google C indentation rules in emacs (https://google-styleguide.googlecode.com/svn/trunk/google-c-style.el)
and it puts 4 spaces indent there, like:

```c
static etimeint_t ecma_DayFromYear(etimeint_t y) {
  return 365 * (y-1970) + floor((y-1969)/4) - floor((y-1901)/100) +
      floor((y-1601)/400);
}
```

I didn't notice much conflicts with cpq's code so I guess it's fine.

alashkin:
I think, the only way for me to survive is to setup C indentation rules for vim.... =)
Add/delete spaces.... Looks like I'll finish tomorrow morning )

mmikulicic:
cpq shared his vim conf:
https://slack-files.com/T02NAP6QJ-F02T2FME1-3468d36255

mmikulicic:
np, no rush

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

alashkin:
No... I'm going to complete today...

src/date.c
alashkin: How? With 4 spaces or by paren?

mmikulicic: paren (i.e. INVALID_TIME on second line starts at same column as INVALID_TIME of the previous line)

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.
BTW, why?
As for me, (res <= 2) looks more "boolean"

src/date.c
alashkin: For me - it is not working with array, it is pointer operation
>> int len = d_tptodatestr(tp, buf, addtz);
>> *(buf + len) = ' ';
>>return d_tptotimestr(tp, buf + len + 1, addtz) + len + 1;

I use (buf + len) as a start point - and as result I wrote *(buf+len) instead of buf[len]

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

src/date.c
alashkin: Done.

alashkin:
>>Awesome! We have dates and 35.something % of ecma coverage!

Actually, this "something" is 85, so ecma coverage is almost 36% )

alashkin:
Marco,
I walked through the style and fix what I found.
(but honestly, I lost freshness of vision already ).

PTAL

alashkin:
Additionally run cpplint and fix several style notes

mmikulicic:
Epic PR, LGTM!

alashkin:
Amen!

PR #263, opened 2015-02-17
Fix s_substr

mmikulicic:
We cannot pass a buffer backed by a string mbuf to
v7_create_tring. For now let's create a temporary buffer.


mmikulicic:
Superseded by #265

PR #286, opened 2015-02-23
Free all allocated ASTs

mmikulicic:
V7 initialization invokes v7_exec. Record all allocated ASTs in order to
free them in v7_destroy


cpq:
Alternative way would be to add `struct ast::next` pointer and link all allocated ASTs into a list - but using mbuf is good too. Arguably less RAM, more reallocs.

Issue #281, opened 2015-02-23
undefined date references


vrz:
date.c:158: undefined reference to `localtime_r'
date.c:368: undefined reference to `getdate'

alashkin:
Hello,
Could you specify your OS & compiler?

vrz:
Tried on WinXP & Win7.
Compiler - GCC(MinGW).

Пользователь Alexander Alashkin <notifications@github.com> писал:

>Hello,CouldyouspecifyyourOScompiler?ReplytothisemaildirectlyorviewitonGitHub.

PR #292, opened 2015-02-26
MinGW support

alashkin:
MinGW support
Closes https://github.com/cesanta/v7/issues/281

### DIFF
--- a/src/date.c
+++ b/src/date.c
@@ -15,9 +15,8 @@
 
 #ifdef __APPLE__
 int64_t strtoll(const char *, char **, int);
-#else
+#elif !defined(_WIN32)
 extern long timezone;
-extern struct tm *getdate(const char *);
 #endif
 
 typedef double etime_t; /* double is suitable type for ECMA time */
@@ -153,9 +152,12 @@ static int ecma_WeekDay(etime_t t) {
 
 static int ecma_DaylightSavingTA(etime_t t) {
   time_t time = t / 1000;
-  struct tm tm;
-  memset(&tm, 0, sizeof(t));
-  localtime_r(&time, &tm);
+  /*
+   * Win32 doesn't have locatime_r
+   * nixes don't have localtime_s
+   * as result using localtime
+   */
+  struct tm tm = *localtime(&time);
   if(tm.tm_isdst > 0) {
     return msPerHour;
   } else {
@@ -352,7 +354,7 @@ static int d_parsedatestr(const char *jstr, size_t len, struct timeparts *tp,
   memset(tp, 0, sizeof(*tp));
   *tz = NO_TZ;
 
-  /* #1: trying toISOSrting() format */
+  /* trying toISOSrting() format */
   {
     const char *frmISOString = " %d-%02d-%02dT%02d:%02d:%02d.%03dZ";
     res = sscanf(str, frmISOString, &tp->year, &tp->month, &tp->day,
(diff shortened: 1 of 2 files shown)


alashkin:
It depends.
p1. Under, for instance,  Minix this function is able to give us +10 additional date formats out-of-box.
p2. It provides workaround if someone wants to have specific date format.
Generally, I added this call to have second point.

cpq:
That's what I am afraid of. If someone writes a script under OS A, it might not work on OS B because of gettdate() call. I suggest to remove it - we'll avoid inconsistencies, and improve portability.

alashkin:
Done,
PTAL, pls

src/date.c
```diff
@@ -363,22 +365,7 @@ static int d_parsedatestr(const char *jstr, size_t len, struct timeparts *tp,
...
-
-      return 1;
-    }
-  }
-
-  /* #3: trying toString()/toUTCString()/toDateFormat() formats */
+  /* trying toString()/toUTCString()/toDateFormat() formats */
   {
     char month[4];
     const char *frmString = " %03*s %03s %02d %d %02d:%02d:%02d %03s%d";
```
mmikulicic: BTW, what is this `%03*s` ? AFAIK if you use `%*s` to ignore a string, the string "precision" doesn't make sense.

alashkin: That is parsing of toXXXString() format.
%03*s skips DOW. If string doesn't starts rom DOW - we can't assume what is it.

mmikulicic: Sorry, bad phrasing; I meant: any reason why this should't be `%*s` ?

(I noticed compiler warnings when running with address sanitizer:

```
==17760==WARNING: unexpected format specifier in scanf interceptor: %03*
==17760==WARNING: unexpected format specifier in scanf interceptor: %03*
==17760==WARNING: unexpected format specifier in scanf interceptor: %03*
==17760==WARNING: unexpected format specifier in scanf interceptor: %03*
==17760==WARNING: unexpected format specifier in scanf interceptor: %03*
```
)

alashkin: Possible it is implementation specific?
I've checked this with clang.
%03*s skips ONLY 3 characters, while %*s skips all before whitespace.

alashkin:
I meant clang on OS X

src/date.c
alashkin: I meant clang on OS X

mmikulicic: I tried with `Apple LLVM version 6.0 (clang-600.0.54) (based on LLVM 3.5svn)`
and `clang version 3.5.0 (tags/RELEASE_350/final)`

Anyway, I get the warnings from the compiler but the actual libsystem sscanf implementation does handle it.

However I'm not sure it works as intended anyway. Namely, when you use `%03*s` it will consume 3 chars and then leave whatever is left to be consumed as the next token, without affecting the actual sscanf return value, so you cannot use this in order to ensure that a three letter DOW token is present.

For example, by using `%03*s` it will actually parse `IncDec 01 2015 00:00:00` as `Tue Dec ...`,
while by having `%*s` it correctly fail to parse that, but it will accept `Foooo Dec 01 2015 00:00:00`.
On the other hand, the current `%03*s` format also does accept DOWs shorter than 3 letters, which is suboptimal as well.

Probably a better way of handling this is:

```c
    const char *frmString = " %*s%n %03s %02d %d %02d:%02d:%02d %03s%d";
    res = sscanf(str, frmString, &dowlen, month, &tp->day, &tp->year,
                 &tp->hour, &tp->min, &tp->sec, gmt, tz);
```

and check that `dowlen == 3`.

Anyway I don't think this is very important, I'm more concerned about becoming warning blind.

Here the compiler doesn't warn via `-Wformat-invalid-specifier`, because the format string is not a literal parameter of `sscanf`, but ASAN issues a runtime warning.

mmikulicic: Suggestion implemented in PR #293.

PR #296, opened 2015-02-26
Embed JS stdlib code

cpq:
Version without separate file

### DIFF
--- a/src/gc.h
+++ b/src/gc.h
@@ -9,6 +9,8 @@
 #include "internal.h"
 #include "vm.h"
 
+#define V7_DISABLE_GC
+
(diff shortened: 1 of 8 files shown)


src/gc.h
```diff
@@ -9,6 +9,8 @@
 #include "internal.h"
 #include "vm.h"
 
+#define V7_DISABLE_GC
```
mmikulicic: if you rebase against master, would this still be necessary?

mmikulicic:
LGTM

mmikulicic:
Did you merge it without first trying out if it works now with GC enabled?

PR #298, opened 2015-02-27
Restore build on MSVC6

cpq:
Re-enable GC.
Fix short strings behavior on 32-bit archs (strings were truncated because of the implicit pointer conversion).

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -172,8 +172,8 @@ V7_PRIVATE void gc_mark(struct v7 *v7, val_t v) {
 
 static void gc_dump_arena_stats(const char *msg, struct gc_arena *a) {
   if (a->verbose) {
-    fprintf(stderr, "%s: total allocations %lu, max %lu, alive %u\n", msg,
-            (unsigned long) a->allocations, a->size, a->alive);
+    fprintf(stderr, "%s: total allocations %lu, max %lu, alive %lu\n", msg,
+            a->allocations, a->size, a->alive);
   }
 }
 
--- a/src/gc.h
+++ b/src/gc.h
@@ -9,7 +9,6 @@
 #include "internal.h"
 #include "vm.h"
 
-
 /* Disable GC on 32-bit platform for now */
 #if ULONG_MAX == 4294967295
 #define V7_DISABLE_GC
@@ -28,8 +27,12 @@ struct gc_cell {
   uintptr_t word;
 };
 
+#ifdef _WIN32
+#define GC_TMP_FRAME(v) struct gc_tmp_frame v = new_tmp_frame(v7);
+#else
 #define GC_TMP_FRAME(v) __attribute__((cleanup(tmp_frame_cleanup), unused)) \
   struct gc_tmp_frame v = new_tmp_frame(v7);
+#endif
 
 #if defined(__cplusplus)
 extern "C" {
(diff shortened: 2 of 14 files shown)


src/gc.h
```diff
@@ -11,20 +11,24 @@
 
 #define V7_DISABLE_GC
 
-#define MARK(p) (* (uintptr_t *) &(p) |= 1)
+#define MARK(p) (* (unsigned long *) &(p) |= 1)
```
mmikulicic: The GC works by tagging the first pointers of a structure, not val_t.

```c
struct v7_object {
  /* First HIDDEN property in a chain is an internal object value */
  struct v7_property *properties;
  struct v7_object *prototype;
};
```

With this change when you mark an object, it will set the LSB of the object's prototype, which means that it will be traversed and it will crash.

src/gc.h
```diff
@@ -11,20 +11,24 @@
...
-#define MARKED(p) ((uintptr_t) (p) & 1)
+#define MARKED(p) (((unsigned long) (void *) (p)) & 1)
 
 struct gc_tmp_frame {
   struct v7 *v7;
   size_t pos;
 };
 
+#ifdef _WIN32
+#define GC_TMP_FRAME(v) struct gc_tmp_frame v = new_tmp_frame(v7);
```
mmikulicic: This is not a functional replacement.

We should replace this with some nop code and disable GC on windows until we handle tmp frame cleanup manually in all functions.

src/vm.c
```diff
@@ -790,7 +790,8 @@ v7_val_t v7_create_string(struct v7 *v7, const char *p, size_t len, int own) {
     embed_string(m, m->len, (char *) &p, sizeof(p));
   }
 
-  return v7_pointer_to_value((void *) offset) | tag;
+  /* NOTE(lsm): don't use v7_pointer_to_value, 32-bit ptrs will truncate */
+  return (offset & ~V7_TAG_MASK) | tag;
```
mmikulicic: the approach with unions in go/typedef-val_t will make this kind of stuff easier to write and maintain.

src/gc.h
cpq: Got it. Restored the old logic with uintptr_t, and typedefed uintptr_t to unsigned long on Win32.
Keep GC disabled coz unit test still coredumps.

src/gc.c
```diff
@@ -133,8 +133,7 @@ void gc_sweep(struct gc_arena *a, size_t start) {
   for (cur = a->base + (start * a->cell_size);
        cur < (a->base + (a->size * a->cell_size));
        cur += a->cell_size) {
-    uintptr_t it = (* (uintptr_t *) cur);
-    if (it & 1) {
+    if (MARKED(cur)) {
```
mmikulicic: crash when GC is enabled is caused by this change.

`MARKED(x)` checks whether the pointer `x` is marked. 

`cur` is a pointer to a cell; the cell is marked if the first pointer sized word is marked:

```c
if (MARKED((* (uintptr_t *) cur))) {
```

mmikulicic: (addressed in PR #301)

src/gc.h
cpq: I've done manual cleanup in vm.c, but it is too much stuff in interpreter.c - it begs a separate PR.

mmikulicic: Do you want me to take over this branch and resolve conflicts so I can make a pass at removing this GCC extension use?

cpq: You can take over - but I think the easier thing is to merge this one and
remove gcc extension in a separate PR

PR #307, opened 2015-02-27
Add Array.prototype.{lastIndexOf,map}

cpq:
Bumps coverage to 39.45%


src/stdlib.c
```diff
@@ -239,7 +239,27 @@ Array.prototype.indexOf = function(a, b) {
...
+  if (!b || b < 0 || b >= this.length) b = this.length - 1;
+  for (var i = b; i >= 0; i--) {
+    if (this[i] == a) {
+      return i;
+    }
+  }
+  return -1;
+};
+
+Array.prototype.xmap = function(a, b) {
```
mmikulicic: dead code?

Should we use this one instead of the one implemented in C?

cpq: Dead code removed.
JS implementation is wrong. If JS array has gaps (e.g. a = [1]; a[100] = 2;), then "for" loop that iterates over includes all indices, whereas only existing ones should be used.
ptal

mmikulicic: Wouldn't 'for in' do the job?

Anyway C impl is fine.

src/array.c
```diff
@@ -292,6 +292,30 @@ static val_t Array_splice(struct v7 *v7, val_t this_obj, val_t args) {
...
+  val_t params, el, res = v7_create_array(v7);
+  struct v7_property *p;
+
+  if (v7_is_undefined(arg1)) {
+    arg1 = this_obj;
+  }
+
+  for (p = v7_to_object(this_obj)->properties; p != NULL; p = p->next) {
+    size_t n;
+    const char *name = v7_to_string(v7, &p->name, &n);
```
mmikulicic: Name pointer can be relocated by any allocation happening below.

mmikulicic:
LGTM, modulo position of call to to_string

cpq:
Made a change in
https://github.com/cesanta/v7/commit/9603d5bb0afd2215729b88a3c4eeac2f4af41e06

PR #310, opened 2015-02-28
re-newed String & RegExp

vrz:
still not work setter for lastIndex

need to correct String literal parsing. I'll do it later


cpq:
Please rebase against the current head

src/regex.c
```diff
@@ -5,150 +5,161 @@
...
+  int flags = 0;
+  val_t r = i_value_of(v7, this_obj);
 
-V7_PRIVATE void Regex_ignoreCase(struct v7_val *this_obj, struct v7_val *arg,
-                                 struct v7_val *result) {
-  if (NULL == result || arg) return;
-  v7_init_bool(result, this_obj->fl.fl.re_flags & SLRE_FLAG_I);
+  (void)args;
+  if (v7_is_regexp(r)) {
+    struct v7_regexp *rp = v7_to_pointer(r);
```
mmikulicic: Doesn't compile as C++ : 

```
../src/regex.c:46:23: error: cannot initialize a variable of type 'struct v7_regexp *' with an rvalue of type 'void *'
    struct v7_regexp *rp = v7_to_pointer(r);
```

(Same thing below)

src/vm.c
```diff
@@ -742,11 +753,13 @@ V7_PRIVATE size_t unescape(const char *s, size_t len, char *to) {
 }
 
 /* Insert a string into mbuf at specified offset */
-V7_PRIVATE void embed_string(struct mbuf *m, size_t offset, const char *p,
+V7_PRIVATE long embed_string(struct mbuf *m, size_t offset, const char *p,
```
mmikulicic: function ends without returning a value.

Why `long` ?

src/vm.c
```diff
@@ -774,11 +787,11 @@ v7_val_t v7_create_string(struct v7 *v7, const char *p, size_t len, int own) {
     s[-1] = len;
     tag = V7_TAG_STRING_I;
   } else if (own) {
-    embed_string(m, m->len, p, len);
+    if(embed_string(m, m->len, p, len) < 0) throw_exception(v7, "Error", "Sintax error");
```
mmikulicic: the first string parameter of `throw_exception` is the error class, and we have "SyntaxError" among valid error types:

`throw_exception(v7, "SyntaxError", "Invalid string")` or something like that

mmikulicic: style: spaces between `if` and `(`

mmikulicic:
Some tests are failing:


`Fail on line 1672: [unescape("\\\n", 2, buf) == 1]`

mmikulicic:
FYI, no need to make another PR. If you force push on your branch diff
comment history is preserved

PR #313, opened 2015-03-01
Change for..in iteration order

cpq:


mmikulicic:
I thought the JS spec mandates that for .. in iterates in the same order as what returned from `Object.keys`, not necessarily the order the elements where created/defined.

mmikulicic:
http://www.ecma-international.org/ecma-262/5.1/#sec-12.6.4

```The mechanics and order of enumerating the properties (step 6.a in the first algorithm, step 7.a in the second) is not specified. ```

http://www.ecma-international.org/ecma-262/5.1/#sec-15.2.3.14

```If an implementation defines a specific order of enumeration for the for-in statement, that same enumeration order must be used in step 5 of this algorithm.```

PR #314, opened 2015-03-01
Fix mingw build

cpq:
Rebased against the master


cpq:
Rebased against the master, ptal

PR #315, opened 2015-03-01
Add Array.prototype.reduce

cpq:
Added Date test 2068 to ignore list.


cpq:
Rebased against master. PTAL

src/stdlib.c
```diff
@@ -251,6 +251,23 @@ Array.prototype.lastIndexOf = function(a, b) {
   return -1;
 };
 ));
+
+  v7_exec(v7, &res, STRINGIFY(
+    Array.prototype.reduce = function(a, b) {
+      var f = 0; var r = b; var t = this;
+      if (typeof(a) != 'function') {
+        throw new TypeError('oops');
```
mmikulicic: `a + 'is not a function'`

cpq: Done.
Also, increased a limit in `i_populate_local_vars()` to make jquery compile

src/stdlib.c
```diff
@@ -251,6 +251,23 @@ Array.prototype.lastIndexOf = function(a, b) {
   return -1;
 };
 ));
+
+  v7_exec(v7, &res, STRINGIFY(
+    Array.prototype.reduce = function(a, b) {
+      var f = 0; var r = b; var t = this;
```
mmikulicic: let's just mutate the argument `b`.
Let's also use `this`. Minification doesn't pay off here. 13 bytes for decl to save 9 bytes. `this` is already passed, `t` takes extra space on the environment and AST-code is longer (`AST_THIS` is a special opcode)

mmikulicic:
LGTM, modulo minor inline comment

src/stdlib.c
cpq: Done

PR #318, opened 2015-03-02
Implement Array.prototype.pop() in JS

cpq:


src/stdlib.c
```diff
@@ -231,26 +231,33 @@ static void init_js_stdlib(struct v7 *v7) {
...
-      return i;
-    }
-  }
-  return -1;
-};
-));
+    Array.prototype.indexOf = function(a, b) {
+      if (!b || b < 0) b = 0;
+      for (var i = b; i < this.length; i++) {
+        if (this[i] == a) {
```
mmikulicic: noticed now: strict equality

src/stdlib.c
```diff
@@ -231,26 +231,33 @@ static void init_js_stdlib(struct v7 *v7) {
...
+        }
+      }
+      return -1;
+    };));
+
+  v7_exec(v7, &res, STRINGIFY(
+    Array.prototype.lastIndexOf = function(a, b) {
+      if (!b || b < 0 || b >= this.length) b = this.length - 1;
+      for (var i = b; i >= 0; i--) {
+        if (this[i] == a) {
```
mmikulicic: same here

src/stdlib.c
```diff
@@ -231,26 +231,33 @@ static void init_js_stdlib(struct v7 *v7) {
...
+        }
+      }
+      return -1;
+    };
+    ));
+
+  v7_exec(v7, &res, STRINGIFY(
+    Array.prototype.pop = function(a, b) {
+      var i = this.length - 1;
+      return this.splice(i, 1)[0];
```
mmikulicic: popping an empty array should return `undefined`, not an empty array.

cpq: It is undefined, isn't it? [][0] returns undefined.

mmikulicic: Oh yes, sorry

cpq:
Addressed strict equality comments, PTAL

Issue #327, opened 2015-03-02
v7_to_string() returns string pointers which cannot be passed to routines that expects null terminated strings


kkutsner:
I see that V7 strings are not null-terminated by design...

A lot of C libraries does not support API to work with non-nullterminated strings. That means the developer need to make a copy of string returned by `v7_to_string()` and add termination symbol before passing it to some library routine. E.g. `sqlite_open()` API expects null terminated strings passed as parameters. The approach when you need to make a temporary copy of string looks not handy to me and leads to unnecessary memory allocations when developing custom interfaces against different 3rdparty libraries.

I believe that adding termination symbol to strings will not seriously increase the memory consumption but it will make the library's API easy to link with other libraries and develop custom interfaces. At least, there could be preprocessor option to add or not to add the termination symbol, but I would not recommend to proceed with such way, because different implemented custom interfaces (released by different developers) may require different options and it will not be possible to link them alltogether.
 
Sorry, if current design is something like a key feature in your library - just close my issue in this case. If not  - I could probably help to you with that.

mmikulicic:
Currently there are 2 main reasons why V7 doesn't store null-terminated strings internally:

1. JS strings can contain 0 bytes in them.

2. Short strings are embedded in the NaN payload of ieee 754 doubles. Especially strings of length 6 cannot have a terminator.

Furthermore, the GC can relocate memory so the validity of these pointers is very short, making a copy often necessary anyways (though not in your example and many similar uses).

That said, the point you make is valid and we should think of a way of avoiding copying strings, especially those of non trivial (4-8 bytes) length.

mmikulicic:
Clarification: We could include a tailing 0 byte in V7 strings, with a modest storage cost, but it would mean that we cannot fit short 6 byte strings in `val_t` values.

A possible solution would be this API:

```c
char *v7_to_string(struct v7 *v7, val_t *v, char buf[7], size_t *sizep);
```

(and copy all 6 byte strings into the user provided buf, null terminate and return it)

but that's perhaps an overkill just to enable us to encode 6 char long strings in `val_t` values given that we can do other things, such as intern strings, use a dictionary of common strings, etc.

The null termination per se is not likely to be a big cost, but the relative cost of loosing compact encoding for a 6 char string (stuff like the 'length' property name, stringyfied numbers from 100000-99999) is quite high (130% more); but I guess usability is quite important as well.

mmikulicic:
[...]

PR #328, opened 2015-03-02
Add V7_TAG_STRING_6

cpq:
No coverage changes.
Inlines strings of length 6.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -42,6 +42,7 @@ enum v7_type val_type(struct v7 *v7, val_t v) {
     case V7_TAG_STRING_I:
     case V7_TAG_STRING_O:
     case V7_TAG_STRING_F:
+    case V7_TAG_STRING_6:
       return V7_TYPE_STRING;
     case V7_TAG_BOOLEAN:
       return V7_TYPE_BOOLEAN;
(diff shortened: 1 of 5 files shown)


mmikulicic:
Let's see how the discussion in #327 about zero termination goes.
It might be necessary to use a special tag to encode STRING_5 and use the space taken by the string length for storing a null byte.

cpq:
I like zero-termination idea. Let's do it as a separate effort.

mmikulicic:
Ok, let's abandon this, and keep the branch since we'll likely do a similar thing for 5-strings

cpq:
Why not merging?
Adding 0-termination is an unrelated effort.
Regardless of it, adding an extra tag to hold a string that exactly fits NAN payload is a good idea.

mmikulicic:
The problem is that if it fits exactly the NaN payload it cannot be null terminated.

So basically this tag will be used for fitting a 5-string plus it's null terminating byte.
Here, I'm assuming that we want to have all strings to be null terminated, that inline strings can have null bytes in them (hence the length will still be encoded in the first byte); but I'd defer to a design discussion to see if those assumptions make sense.

Basically, there are a few choices to be made, and I thought that making them without the burden of caring for another string type would be leaner.

PR #332, opened 2015-03-04
Add V7_TAG_STRING_5

### LINKED ISSUE #327: v7_to_string() returns string pointers which cannot be passed to routines that expects null terminated strings

cpq:
Add special tag for strings that are exactly 5 bytes long.
Fix #327.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -42,6 +42,7 @@ enum v7_type val_type(struct v7 *v7, val_t v) {
     case V7_TAG_STRING_I:
     case V7_TAG_STRING_O:
     case V7_TAG_STRING_F:
+    case V7_TAG_STRING_5:
       return V7_TYPE_STRING;
     case V7_TAG_BOOLEAN:
       return V7_TYPE_BOOLEAN;
@@ -72,7 +73,8 @@ int v7_is_function(val_t v) {
 
 int v7_is_string(val_t v) {
   uint64_t t = v & V7_TAG_MASK;
-  return t == V7_TAG_STRING_I || t == V7_TAG_STRING_F || t == V7_TAG_STRING_O;
+  return t == V7_TAG_STRING_I || t == V7_TAG_STRING_F ||
+    t == V7_TAG_STRING_O || t == V7_TAG_STRING_5;
 }
 
 int v7_is_boolean(val_t v) {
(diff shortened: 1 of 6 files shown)


src/vm.c
```diff
@@ -892,6 +906,7 @@ V7_PRIVATE val_t s_concat(struct v7 *v7, val_t a, val_t b) {
   }
   memcpy(s, a_ptr, a_len);
   memcpy(s + a_len, b_ptr, b_len);
+  s[a_len + b_len] = '\0';
```
mmikulicic: This looks like it's not necessary if len <= 5; could we put it only in the else branch?

mmikulicic:
Looks good, I wonder why the test fails

cpq:
No new failing tests!

src/vm.c
cpq: I'd keep it like that: that way we stress that regardless of the string type, it gets 0 terminated. Also, conditional would generate (arguably) longer and slower code. I've put a comment though.

mmikulicic: sgtm.

w.r.t the conditional, it's already there isn't it ? It'd be just a matter
of moving the statement at the end of the else branch, right?
(anyway it doesn't matter, it would be nano-optimisation)

mmikulicic:
Tested also with MSAN on linux. No new failing tests (there is a pre-existing issue with dates though)

PR #336, opened 2015-03-04
Fix array length size

mmikulicic:
Was overflowing on 32-bit

---

This doesn't really solve the issue of overflows;
we're plagued, but at least it doesn't segfault
and so we can see if other changes are not breaking the m32 build


cpq:
Merged #340 instead

PR #339, opened 2015-03-05
Limit array size to 2^32 

cpq:
Oops disregard this PR.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -737,7 +737,6 @@ V7_PRIVATE long v7_array_length(struct v7 *v7, val_t v) {
   long max = -1, k;
   char *end;
 
-  (void) v7;
   if (!v7_is_object(v)) {
     return -1;
   }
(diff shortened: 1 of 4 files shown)


src/vm.c
```diff
@@ -746,10 +745,12 @@ V7_PRIVATE long v7_array_length(struct v7 *v7, val_t v) {
     size_t n;
     const char *s = v7_to_string(v7, &p->name, &n);
     k = strtol(s, &end, 10);
-    if (end != s && k > max) {
+    /* Array length could not be more then 2^32 */
+    if (end > s && k > max && k < 4294967295L) {
```
mmikulicic: `long` is 31 bit quantity on m32.

cpq: yeah, unsigned long is a way to go. it is guaranteed to be at least 32 bit. i've killed this PR, let's go ahead w/yours

PR #340, opened 2015-03-05
Limit array size to 2^32 

cpq:
Hi @alashkin, please make a review! Since @mmikulicic is off today, let this PR take over PR #336

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -732,25 +732,26 @@ V7_PRIVATE val_t v7_property_value(struct v7 *v7, val_t obj,
   return p->value;
 }
 
-V7_PRIVATE long v7_array_length(struct v7 *v7, val_t v) {
+V7_PRIVATE unsigned long v7_array_length(struct v7 *v7, val_t v) {
   struct v7_property *p;
-  long max = -1, k;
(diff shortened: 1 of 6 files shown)


src/vm.c
```diff
@@ -732,12 +732,11 @@ V7_PRIVATE val_t v7_property_value(struct v7 *v7, val_t obj,
...
+V7_PRIVATE unsigned long v7_array_length(struct v7 *v7, val_t v) {
   struct v7_property *p;
-  long max = -1, k;
+  unsigned long max = 0, k;
   char *end;
 
-  (void) v7;
   if (!v7_is_object(v)) {
     return -1;
   }
```
alashkin: How `return -1` will work with unsigned return type?

alashkin: I mean, it gives 0xFFFFFFF...... and will be lenght number on 32b.

cpq:
PTAL

src/vm.c
cpq: Good catch. Fixed (returning 0).

alashkin:
LGTM

PR #361, opened 2015-03-14
Improve string to number conversion

mmikulicic:
Simplify and minor performance improvements

DIFFBASE=#360


mmikulicic:
Note: this works now because we have null-terminated strings (as of #327)

PR #368, opened 2015-03-15
Improve property string comparison

mmikulicic:
Short strings are compared by val_t value

DIFFBASE=#367

---

@cpq, this doesn't currently buy much (~6% of the whole ecma test run),
it can be improved by passing down the val_t of the key if known (usually it is if
the lookup is performed by the VM, as opposed by the user who passes a char *).


mmikulicic:
PTAL,

It also contains some formatting changes caused by running clang-format.
I wonder if you forgot to run clang-format and the CI build failed to check for misformats, or if we all have slightly different versions of clang-format.

mmikulicic:
PTAL

PR #379, opened 2015-03-19
Refactor C constructors and error objects

mmikulicic:
New'ed C constructors are now called with `this`'s prototype
correctly set to the prototype defined in the C function object
prototype property (for those C constructors that are also function objects).

This makes it simpler to write C constructors and enables reusing the same
C constructor function for a whole inhertiance tree (e.g. error objects)

ECMA coverage up to 50.92 %
Small performance improvement

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/379)

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -5,25 +5,40 @@
 
 #include "internal.h"
 
-V7_PRIVATE void init_error(struct v7 *v7) {
-  val_t v;
-  v7_exec(v7, &v, "function Error(m) {this.message = m}");
-  v7_exec(v7, &v,
-          "function TypeError(m) {this.message = m};"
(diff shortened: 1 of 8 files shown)


alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/379)

Reviewed files:
- src/error.c @ r1
- src/gc.c @ r1
- src/internal.h @ r1
- src/interpreter.c @ r1
- src/number.c @ r1
- src/stdlib.c @ r1
- tests/ecma_report.txt @ r1
- v7.c @ r1





---

alashkin:
LGTM

I'll try it )

PR #380, opened 2015-03-19
Create exceptions with C code

mmikulicic:
It doesn't require exec-ing every time an internal exception is raised

DIFFBASE=#379

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/380)

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -40,11 +40,11 @@ static val_t Array_set_length(struct v7 *v7, val_t this_obj, val_t args) {
   long new_len = arg_long(v7, args, 0, -1);
 
   if (!v7_is_object(this_obj)) {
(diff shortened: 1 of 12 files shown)


alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/380)

Reviewed files:
- src/array.c @ r1
- src/boolean.c @ r1
- src/date.c @ r1
- src/function.c @ r1
- src/internal.h @ r1
- src/interpreter.c @ r1
- src/number.c @ r1
- src/object.c @ r1
- src/regex.c @ r1
- src/string.c @ r1
- src/vm.c @ r1
- v7.c @ r1





---

PR #385, opened 2015-03-19
Small Object.defineProperty enhancement

cpq:
Hi @alashkin , please make a review.
This bumps coverage to 50.94%.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/385)

### DIFF
--- a/src/json.c
+++ b/src/json.c
@@ -16,10 +16,7 @@ static val_t Json_stringify(struct v7 *v7, val_t this_obj, val_t args) {
 
 V7_PRIVATE void init_json(struct v7 *v7) {
   val_t o = v7_create_object(v7);
-  unsigned flags = V7_PROPERTY_DONT_ENUM;
-
-  v7_set_property(v7, o, "stringify", 9, flags,
-                  v7_create_cfunction(Json_stringify));
-  v7_set_property(v7, o, "parse", 5, flags, v7_create_cfunction(Std_eval));
-  v7_set_property(v7, v7->global_object, "JSON", 4, 0, o);
+  set_cfunc_obj_prop(v7, o, "stringify", Json_stringify, 1);
+  set_cfunc_obj_prop(v7, o, "parse", Std_eval, 1);
+  v7_set_property(v7, v7->global_object, "JSON", 4, V7_PROPERTY_DONT_ENUM, o);
 }
--- a/src/object.c
+++ b/src/object.c
@@ -93,21 +93,37 @@ static val_t Obj_getOwnPropertyDescriptor(struct v7 *v7, val_t this_obj,
   return desc;
 }
 
+static void o_set_attr(struct v7 *v7, val_t desc, const char *name, size_t n,
+                       struct v7_property *prop, unsigned int attr) {
+  val_t v = v7_get(v7, desc, name, n);
+  if (v7_is_true(v7, v)) {
+    prop->attributes &= ~attr;
+  } else {
+    prop->attributes |= attr;
+  }
+}
+
 static val_t _Obj_defineProperty(struct v7 *v7, val_t obj, const char *name,
                                  int name_len, val_t desc) {
-  unsigned int flags = 0;
   val_t val = v7_get(v7, desc, "value", 5);
-  if (!v7_is_true(v7, v7_get(v7, desc, "enumerable", 10))) {
-    flags |= V7_PROPERTY_DONT_ENUM;
-  }
-  if (!v7_is_true(v7, v7_get(v7, desc, "writable", 8))) {
-    flags |= V7_PROPERTY_READ_ONLY;
-  }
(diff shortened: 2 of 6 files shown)


alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/385)

Reviewed files:
- src/json.c @ r1
- src/vm.c @ r1
- tests/ecma_report.txt @ r1
- v7.c @ r1





---

<sup>**[src/object.c, line 100 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/385#-Jkq9tl9eBpl7wy_uIvd)** ([raw file](https://github.com/cesanta/v7/blob/1ca855db10433f2b791073e2979c7317fe63014f/src/object.c#L100)):</sup>
This line causes another behavior comparing to previous version:
previous one never clear attribute, as new one does.
So, for example, if object already had V7_PROPERTY_DONT_ENUM previous implementation kept it. New one will erase it.
Doesn't it produce any side-effect?

---

cpq:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/385)







---

<sup>**[src/object.c, line 100 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/385#-Jkq9tl9eBpl7wy_uIvd)** ([raw file](https://github.com/cesanta/v7/blob/1ca855db10433f2b791073e2979c7317fe63014f/src/object.c#L100)):</sup>
Yeah, let's see. The previous code was as follows:

    flags = 0;
    if (!v7_is_true(v7, v7_get(v7, desc, "enumerable", 10))) {
       flags |= V7_PROPERTY_DONT_ENUM
    }
    ...
    v7_set_property(,,, flags, val);

`!v7_is_true` is true when the argument is false or undefined. Meaning, if `enumerable` is either false or not specified at all, `DONT_ENUM` is set. Note that the subsequent call will overwrite the attribute:

    Object.defineProperty(o, "x", {enumerable: true});  // no DONT_ENUM 
    Object.defineProperty(o, "x", {});  // sets DONT_ENUM but should not

New code does not do the override.

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/385)







---

<sup>**[src/object.c, line 100 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/385#-Jkq9tl9eBpl7wy_uIvd)** ([raw file](https://github.com/cesanta/v7/blob/1ca855db10433f2b791073e2979c7317fe63014f/src/object.c#L100)):</sup>
> Object.defineProperty(o, "x", {});  // sets DONT_ENUM but should not

enumerable defaults to false, so DONT_ENUM should be set here.

---

alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/385)

Reviewed files:
- src/vm.h @ r1





---

<sup>**[src/object.c, line 100 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/385#-Jkq9tl9eBpl7wy_uIvd)** ([raw file](https://github.com/cesanta/v7/blob/1ca855db10433f2b791073e2979c7317fe63014f/src/object.c#L100)):</sup>
Marko, 
Wdym?
That's all is fine or not?

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/385)







---

<sup>**[src/object.c, line 100 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/385#-Jkq9tl9eBpl7wy_uIvd)** ([raw file](https://github.com/cesanta/v7/blob/1ca855db10433f2b791073e2979c7317fe63014f/src/object.c#L100)):</sup>
> Object.defineProperty(o, "x", {});  // sets DONT_ENUM but should not
I parsed it as:  "`defineProperty(o, "x", {})` should not set DON_ENUM" and it caught my eye.

The code itself looks good. Our flags are defined as the exact opposite of the ones in the spec, i.e. an enumerable prop has it's DONT_ENUM flag cleared.

Hence `o_set_attr` will set the flag if the named property is false or missing which is fine since missing "enumerable", "configurable", "writable"  properties default to false.

---

alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/385)







---

<sup>**[src/object.c, line 100 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/385#-Jkq9tl9eBpl7wy_uIvd)** ([raw file](https://github.com/cesanta/v7/blob/1ca855db10433f2b791073e2979c7317fe63014f/src/object.c#L100)):</sup>
Ok

---

alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/385)

Reviewed files:
- src/object.c @ r1





---

PR #393, opened 2015-03-22
Fix Array.map and Array.every index type

mmikulicic:
ECMA coverage at 51.75%

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/393)


cpq:
This code:

    ./v7 -e 'a = [1,2,3]; a[99999999] = 2; a.every(function(a,b,c){return a > 0})'

basically freezes this implementation. While iteration over properties works fast. Is there any downside for keeping iteration as is, but insert a check for valid array key, and pass a numeric arg to callback?

mmikulicic:
ok, my intention was to introduce dense arrays and use them to when arrays are small and contiguous. I can make a bigger PR where we support both of them :-)
We need both code paths, because one property per array element is an overkill for many things arrays are used for (e.g. `arguments`)

PR #413, opened 2015-03-25
Fix msvc6 build, floats

mmikulicic:
It also fixes 0/0 on other platforms and reverts 5d30094

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/413)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,8 +54,8 @@ m32_v7: $(TOP_HEADERS) $(TOP_SOURCES) v7.h
 	$(CC) $(TOP_SOURCES) -o v7 -DV7_EXE -DV7_EXPOSE_PRIVATE $(CFLAGS) -m32 -lm
 
 w: v7.c
-	wine cl v7.c /Zi -DV7_EXE -DV7_EXPOSE_PRIVATE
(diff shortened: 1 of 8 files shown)


alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/413)

Reviewed files:
- Makefile @ r1
- scripts/platform.mk @ r1
- src/internal.h @ r1
- src/interpreter.c @ r1
- src/vm.c @ r1
- tests/ecma_report.txt @ r1
- tests/unit_test.c @ r1
- v7.c @ r1





---

<sup>**[scripts/platform.mk, line 20 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/413#-JlJbMnQIr5yFAcfX56Y)** ([raw file](https://github.com/cesanta/v7/blob/549a5edb88abd7dfc3b6c5f84563ee2c9de0972f/scripts/platform.mk#L20)):</sup>
I want to fix sockets on windows, instead of disable it. 

---

mmikulicic:
> I want to fix sockets on windows, instead of disable it.

Cool. When you fix sockets on windows you change the flag in the same
commit, so that the CI build will check it.

As a rule, better disabling tests than broken CI builds.

PR #429, opened 2015-03-28
redesign Function_ctor()

vrz:
[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/429)

### DIFF
--- a/src/function.c
+++ b/src/function.c
@@ -6,42 +6,42 @@
 #include "internal.h"
 
 static val_t Function_ctor(struct v7 *v7, val_t this_obj, val_t args) {
-  long i, n = 0, num_args = v7_array_length(v7, args);
+  long i, num_args = v7_array_length(v7, args);
   size_t size;
-  char buf[200];
   const char *s;
-  val_t param, body, res = v7_create_undefined();
+  struct mbuf m;
+  val_t tmp;
+  enum v7_err ret;
 
-  (void) this_obj;
+  (void)this_obj;
 
-  if (num_args <= 0) return res;
+  if (num_args <= 0) return v7_create_undefined();
 
-  /* TODO(lsm): Constructing function source code here. Optimize this. */
-  n += snprintf(buf + n, sizeof(buf) - n, "%s", "(function(");
+  mbuf_init(&m, 0);
+  mbuf_append(&m, "(function(", 10);
 
   for (i = 0; i < num_args - 1; i++) {
-    param = i_value_of(v7, v7_array_get(v7, args, i));
-    if (v7_is_string(param)) {
-      s = v7_to_string(v7, &param, &size);
-      if (i > 0) {
-        n += snprintf(buf + n, sizeof(buf) - n, "%s", ",");
-      }
-      n += snprintf(buf + n, sizeof(buf) - n, "%.*s", (int) size, s);
+    tmp = i_value_of(v7, v7_array_get(v7, args, i));
+    if (v7_is_string(tmp)) {
+      if (i > 0) mbuf_append(&m, ",", 1);
+      s = v7_to_string(v7, &tmp, &size);
+      mbuf_append(&m, s, size);
     }
   }
-  n += snprintf(buf + n, sizeof(buf) - n, "%s", "){");
-  body = i_value_of(v7, v7_array_get(v7, args, num_args - 1));
-  if (v7_is_string(body)) {
-    s = v7_to_string(v7, &body, &size);
-    n += snprintf(buf + n, sizeof(buf) - n, "%.*s", (int) size, s);
+  mbuf_append(&m, "){", 2);
+  tmp = i_value_of(v7, v7_array_get(v7, args, num_args - 1));
+  if (v7_is_string(tmp)) {
+    s = v7_to_string(v7, &tmp, &size);
+    mbuf_append(&m, s, size);
   }
-  n += snprintf(buf + n, sizeof(buf) - n, "%s", "})");
+  mbuf_append(&m, "})\0", 3);
 
-  if (v7_exec_with(v7, &res, buf, V7_UNDEFINED) != V7_OK) {
+  ret = v7_exec(v7, &tmp, m.buf);
+  mbuf_free(&m);
+  if (ret != V7_OK) {
     throw_exception(v7, SYNTAX_ERROR, "Invalid function body");
   }
-
-  return res;
+  return tmp;
 }
 
 static val_t Function_length(struct v7 *v7, val_t this_obj, val_t args) {
--- a/v7.c
+++ b/v7.c
@@ -12854,42 +12854,42 @@ V7_PRIVATE void init_date(struct v7 *v7) {
 
 
 static val_t Function_ctor(struct v7 *v7, val_t this_obj, val_t args) {
-  long i, n = 0, num_args = v7_array_length(v7, args);
+  long i, num_args = v7_array_length(v7, args);
   size_t size;
-  char buf[200];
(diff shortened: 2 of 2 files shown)


mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)







---

<sup>**[src/function.c, line 34 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-Jlb45amWE2dPnX-Ss8P)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L34)):</sup>
please put braces when the if body is multiline (in this case because wrapped)

---

<sup>**[src/function.c, line 37 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-Jlb4BthEH6pGozmdsp_)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L37)):</sup>
now JS strings (except AST strings for now) are guaranteed to be null terminated

---

<sup>**[src/function.c, line 42 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-Jlb4PcN5JN7GzPf9sig)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L42)):</sup>
why `tmp`? The error message refers to the function body passed as argument, not the `body`local variable.

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)

Reviewed files:
- src/function.c @ r1
- v7.c @ r1

---

This PR is ok, but:

while it looks handy to offload string concatenation to high level V7 string handling functions, it has a few disadvantages:

- complicated variable registering for GC safety
- every concatenation allocates a new string, leaving the old one on the heap as garbage. This introduces unnecessary pressure on the GC, which could be easily avoided by using mbuf_append to keep a dynamic buffer. We could have a helper that renders a string into an mbuf

---

<sup>**[src/function.c, line 13 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-JlbKKeuxzYkoEW4CdBB)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L13)):</sup>
All these values have to registered so that the GC can see them as alive (and relocate pointers during compaction).

```c
struct gc_tmp_frame tf = new_tmp_frame(v7);
tmp_stack_push(&tf, &fn);
tmp_stack_push(&tf, &tmp);
...
tmp_frame_cleanup(&tf);
return ....;
```
Notes:
- when pushing a variable, it should be initialized to something (e.g. undefined)
- throw_exception automatically unwinds the tmp stack as well, so no need to perform cleanup before throwing).

see for example `vm.c:v7_create_function`

---

vrz:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)







---

<sup>**[src/function.c, line 13 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-JlbKKeuxzYkoEW4CdBB)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L13)):</sup>
done

---

<sup>**[src/function.c, line 37 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-Jlb4BthEH6pGozmdsp_)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L37)):</sup>
great! in this case not necessary to make copy of string

---

<sup>**[src/function.c, line 42 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-Jlb4PcN5JN7GzPf9sig)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L42)):</sup>
it heppends just  because i replace word body to tmp

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)

Reviewed files:
- src/function.c @ r2
- v7.c @ r2





---

<sup>**[src/function.c, line 15 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/429#-JleOJNcDHsJPOlAtKkT)** ([raw file](https://github.com/cesanta/v7/blob/455269a4aea9d055d55e60dbe6ffea5ff40f0056/src/function.c#L15)):</sup>
`fn` and `tmp` are uninitialized, if `fn = v7_create_string(...` below triggers a GC, we can potentially mutate touch uninitialized memory (low probability because it has to look like a NaN encoded object pointer, but still).

---

vrz:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)



---

done

---

<sup>**[src/function.c, line 13 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-JlbKKeuxzYkoEW4CdBB)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L13)):</sup>
done

---

<sup>**[src/function.c, line 37 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-Jlb4BthEH6pGozmdsp_)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L37)):</sup>
great! in this case not necessary to make copy of string

---

<sup>**[src/function.c, line 42 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/429#-Jlb4PcN5JN7GzPf9sig)** ([raw file](https://github.com/cesanta/v7/blob/7a0d25abb567d80e16ac6567afee1db521a1b91f/src/function.c#L42)):</sup>
it heppends just  because i replace word body to tmp

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)



---

LGTM, minor comment

---

<sup>**[src/function.c, line 39 \[r3\]](https://reviewable.io:443/reviews/cesanta/v7/429#-JlenoNlVrRWbOK6tl8n)** ([raw file](https://github.com/cesanta/v7/blob/9c8b4f444580316cef4c7cd0589c38c2a978bec8/src/function.c#L39)):</sup>
you have to free the mbuf here as well (or save v7_exec return value and check after free).

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)

Reviewed files:
- src/function.c @ r3
- v7.c @ r3





---

vrz:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)







---

<sup>**[src/function.c, line 39 \[r3\]](https://reviewable.io:443/reviews/cesanta/v7/429#-JlenoNlVrRWbOK6tl8n)** ([raw file](https://github.com/cesanta/v7/blob/9c8b4f444580316cef4c7cd0589c38c2a978bec8/src/function.c#L39)):</sup>
done

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/429)

Reviewed files:
- src/function.c @ r4
- v7.c @ r4





---

PR #433, opened 2015-03-30
OS interface: renaming File interface

gelraen:
[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/433)

### DIFF
--- a/src/os.c
+++ b/src/os.c
@@ -8,10 +8,11 @@
 #ifndef V7_NO_FS
 /*
  * File interface: a wrappers around open(), close(), read(), write().
- * OS.open(path, flags) -> fd.
+ * OS.open(path, flags[, mode]) -> fd.
  * OS.close(fd) -> undefined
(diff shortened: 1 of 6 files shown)


mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/433)

Reviewed files:
- src/os.c @ r1
- src/sources.mk @ r1
- src/stdlib.c @ r1
- src/vm.h @ r1
- tests/unit_test.c @ r1
- v7.c @ r1

---

number of args is indeed useless especially for our own APIs. ECMA tests do exercise the length property of the standard functions but I don't know how much real world code actually depends on that. It might be useful for JS functions so they can easily slice out the "rest" part of the actuals in the `arguments` array, but we probably don't need that sugar for cfunctions. This is more an artefact of our attempt to leverage the ECMA test suite to figure out what big parts of the spec we are not covering, and some tests were stuck at checking `length` and we didn't want to patch the tests instead.

---

PR #435, opened 2015-03-31
Fix missing recording GC temp root

mmikulicic:
Found while trying some sunspider benchmark

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/435)


alashkin:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/435)

Reviewed files:
- src/function.c @ r1
- src/interpreter.c @ r1
- v7.c @ r1





---

PR #440, opened 2015-03-31
Add OS.rename and OS.uname methods.

gelraen:
[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/440)

### DIFF
--- a/src/internal.h
+++ b/src/internal.h
@@ -21,6 +21,25 @@
  */
 #define V7_BROKEN_NAN
 
+/* Define platform info for systems that don't have POSIX uname() function. */
+#ifdef V7_NO_POSIX
+#ifndef V7_UNAME_SYSNAME
+#define V7_UNAME_SYSNAME "Unspecified OS"
+#endif
+#ifndef V7_UNAME_NODENAME
+#define V7_UNAME_NODENAME "Unspecified hostname"
+#endif
+#ifndef V7_UNAME_RELEASE
+#define V7_UNAME_RELEASE "Unspecified release"
+#endif
+#ifndef V7_UNAME_VERSION
+#define V7_UNAME_VERSION "Unspecified version"
+#endif
+#ifndef V7_UNAME_MACHINE
+#define V7_UNAME_MACHINE "Unspecified architecture"
+#endif
+#endif
+
 #ifdef __GNUC__
 #define NORETURN __attribute__((noreturn))
 #else
@@ -36,7 +55,9 @@
 #define _POSIX_C_SOURCE 200809L
 
 #include <sys/stat.h>
+#ifndef V7_NO_POSIX
 #include <sys/utsname.h>
+#endif
 #include <assert.h>
 #include <ctype.h>
 #include <errno.h>
--- a/src/internal.h
+++ b/src/internal.h
@@ -36,6 +36,7 @@
 #define _POSIX_C_SOURCE 200809L
 
 #include <sys/stat.h>
+#include <sys/utsname.h>
 #include <assert.h>
 #include <ctype.h>
 #include <errno.h>
--- a/src/os.c
+++ b/src/os.c
@@ -86,6 +86,44 @@ static val_t OS_open(struct v7 *v7, val_t this_obj, val_t args) {
 
   return v7_create_number(fd);
 }
+
+static val_t OS_rename(struct v7 *v7, val_t this_obj, val_t args) {
+  val_t arg0 = v7_array_get(v7, args, 0);
+  val_t arg1 = v7_array_get(v7, args, 1);
+  int res = -1;
+
+  (void) this_obj;
+  if (v7_is_string(arg0) && v7_is_string(arg1)) {
+    size_t n1, n2;
+    const char *from = v7_to_string(v7, &arg0, &n1);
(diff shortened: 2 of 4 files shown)


gelraen:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

Do I need to put V7_NO_POSIX knob elsewhere? Is there a better name for it?

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

No idea on both questions.

---

<sup>**[v7.c, line 14348 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/440#-JlkoGrnJScGSGyqqY1s)** ([raw file](https://github.com/cesanta/v7/blob/96c11ce19ac1e08d182d6ad83ce56151eedd0651/v7.c#L14348)):</sup>
should we fake it when it's not there (e.g. let the user provide some preprocessor macro with the values) ?

---

<sup>**[v7.c, line 14357 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/440#-Jlkod3LXn3cpxdl3g4A)** ([raw file](https://github.com/cesanta/v7/blob/96c11ce19ac1e08d182d6ad83ce56151eedd0651/v7.c#L14357)):</sup>
what are the possible failure modes of`uname`  ?

If they are not common, should't we just throw an exception here?

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)

Reviewed files:
- src/internal.h @ r1
- src/os.c @ r1
- tests/unit_test.c @ r1
- v7.c @ r1





---

gelraen:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)







---

<sup>**[v7.c, line 14348 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/440#-JlkoGrnJScGSGyqqY1s)** ([raw file](https://github.com/cesanta/v7/blob/96c11ce19ac1e08d182d6ad83ce56151eedd0651/v7.c#L14348)):</sup>
Done.

---

<sup>**[v7.c, line 14357 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/440#-Jlkod3LXn3cpxdl3g4A)** ([raw file](https://github.com/cesanta/v7/blob/96c11ce19ac1e08d182d6ad83ce56151eedd0651/v7.c#L14357)):</sup>
I'm not aware of any and http://pubs.opengroup.org/onlinepubs/9699919799/functions/uname.html doesn't tell much about that.

What is the criteria for choosing between returning an error and throwing an exception?

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)

Reviewed files:
- src/internal.h @ r2
- src/os.c @ r2
- v7.c @ r2





---

<sup>**[v7.c, line 14357 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/440#-Jlkod3LXn3cpxdl3g4A)** ([raw file](https://github.com/cesanta/v7/blob/96c11ce19ac1e08d182d6ad83ce56151eedd0651/v7.c#L14357)):</sup>
Currently we opted for error values, e.g. in the Socket API, because IO is expected to fail and errors might by recoverable and you're expected to handle it without to much boilerplate and chances for leaking open file descriptors (as we might not want to implement finalizers).

On the other hand, we do use exceptions for all kind of JS runtime errors, and an obscure system error like this might make sense to treat like "hey the runtime is broken, ah btw, it's because of _this_, and _here_".

---

cpq:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)

Reviewed files:
- src/internal.h @ r2
- src/os.c @ r2
- tests/unit_test.c @ r1
- v7.c @ r2





---

cpq:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

On non-POSIX systems, I'd rather not have uname() at all. Ditto for rename().

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

Is there a better way to have multiplatform JS scripts ?
Shall we, ahem ahem, add support for conditional compilation to JS (in order to reduce the size of the AST as well) ?

---

cpq:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

I don't think so.
On embedded, I expect very small & simple JS code, akin to small arduino sketches. On those I'd rather see absent API then stubs, to save space.

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

ok, SGTM

---

gelraen:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

Umm, so it was merged anyway?

---

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/440)



---

@cpq trigger happy ? :-)

---

Issue #442, opened 2015-04-01
Build V7 on Arduino UNO


cpq:
- [ ] Wrap each method in `ifdef`
- [ ] Define a set of pre-defined configurations, e.g. V7_ANOREXIC, V7_THIN, V7_FIT, V7_SLIGHTLY_FAT, V7_BLOATED, V7_COMPLIANT

PR #446, opened 2015-04-03
Introduce feature selection machinery and cut Date into pieces.

rojer:
This is the first step of factoring V7 into distinct "features".
This commit introduces machinery and applies it to Date.
The diff looks scary, but rearranging is necessary to account for dependencies between features.
With this rearrangement, each feature is individually selectable (tested manually for now, might want to do it automatically in the future).

#442 

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/446)


cpq:
Please squash this PR into a single commit

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

mmikulicic:
Reviewed files:
- Makefile @ r1
- scripts/gen-features-full.pl @ r1
- src/date.c @ r1
- src/features_full.h @ r1
- src/features_medium.h @ r1
- src/features_minimal.h @ r1
- src/internal.h @ r1
- src/sources.mk @ r1
- v7.c @ r1

---

<sup>**[Makefile, line 40 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyURIcS-D3mQ2ee3qD)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/Makefile#L40)):</sup>
I'm not sure I like this kind of proliferation of build targets and generated artifacts:

- requires adding those artefacts to `.gitignore` (todo in this PR)
- make clean target (todo in this PR)
- it doesn't play nicely with your history. I usually invoke `./v7 tests/tests/ecma_driver.js test.js` or some other more complicated stuff.


Should we just use this as a phony target that:

```
rm v7 -f && $(MAKE) v7 CFLAGS_EXTRA="-D V7_BUILD_PROFILE=2 $(CFLAGS_EXTRA)"
```

?

(this is more consistent with targets like `m32_v7`, albeit spelled out differently)

Or should we generate all our binaries in some `build` directory and have a symlink `v7` pointing to the latest build so one can choose (and clean and gitginore just need to know about that dir) ?

---

<sup>**[src/date.c, line 789 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyPKk9i1eT4DpgJGg_)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/date.c#L789)):</sup>
this comment now looks misplaced.
We could even do without "section marker" comments.

---

<sup>**[src/features_full.h, line 12 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyPjrvHIXc0z62s6zr)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/features_full.h#L12)):</sup>
1. How can I disable for example `toJSON` ?
Once a preprocessor var is defined it cannot be undefined from the cmdline (`-U` reverts only `-D` specified from cmdline).

2. Image have a cmdline like `make v7 CFLAGS_EXTRA='-DV7_BUILD_PROFILE_MINIMAL -DV7_ENABLE_Date__now"`. Now something doesn't work, and you want to temporarily switch to a higher profile. You have to remove the -DV7_ENABLE_Data__now from your flags because otherwise it will complain with `warning: 'V7_ENABLE_Data__now' macro redefined`

Proposed solution:

```c
#ifndef V7_DISABLE_Date_now
#ifndef V7_ENABLE_Date__now
#define V7_ENABLE_Date__now 1
#endif
#else 
#ifdef V7_ENABLE_Date__now
#undef V7_ENABLE_Date__now 
#endif
#endif
```

Properties:

- Disables are meant to be defined only via external flags. 
- Disables win over enables.
- External enables override the profile (because the profile doesn't )

This is fine for generated code, but perhaps it prompts us to also generate the headers for manually selected profile (out of simpler profile definition files)

---

<sup>**[src/features_full.h, line 14 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlySvyYOgoJOSo89DTv)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/features_full.h#L14)):</sup>
Does defining to `1` have some better properties as opposed to just defining without any value?

---

<sup>**[src/features_minimal.h, line 3 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyW_wxL1IQN2cestxU)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/features_minimal.h#L3)):</sup>
How should we call the feature that decides wether we have a Date object at all ?
`V7_ENABLE_Date` ?

---

<sup>**[src/internal.h, line 81 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyQTYCuh7GIhyBtHhp)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/internal.h#L81)):</sup>
let's just include all of them, the guards are in the files themselves (required for amalgamation anyway). This just generates nops in the build artifact:

```c
#if V7_BUILD_PROFILE == V7_BUILD_PROFILE_MINIMAL
#elif V7_BUILD_PROFILE == V7_BUILD_PROFILE_MEDIUM
#else
#define V7_BUILD_PROFILE V7_BUILD_PROFILE_FULL
#endif
```

---

<sup>**[src/sources.mk, line 7 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyTC0FloTt3UIR8X-D)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/sources.mk#L7)):</sup>
shouldn't we put those at the beginning in case we have to put a guard around some stuff in the headers (e.g. disabling the parser we can also disable chunks of the `v7` structure` and disable some type definitions which can help us spot places that depend on the parser) ?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

rojer:
done

---



---

<sup>**[Makefile, line 40 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyURIcS-D3mQ2ee3qD)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/Makefile#L40)):</sup>
i see your point about having one binary, OTOH while developing it's useful to build all the flavors, to make sure they all build and then be able to compare and test all of them.
for example, here's what i do to test profiles:
$ make -j3 v7 medium_v7 minimal_v7
$ for v7 in v7 medium_v7 minimal_v7; do echo === $v7; ./$v7 -e 'print(Date.parse); print(Date.UTC); print(Date.now);'; done

---

<sup>**[src/date.c, line 789 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyPKk9i1eT4DpgJGg_)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/date.c#L789)):</sup>
Done.

---

<sup>**[src/features_full.h, line 12 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyPjrvHIXc0z62s6zr)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/features_full.h#L12)):</sup>
well, the proposed solution uses 9 lines to define a switch for just one feature. profiles will be unwieldy.
before we go with that, i want to make sure we really want ability to disable individual features form the command line.
do we? this was never mentioned until now.

---

<sup>**[src/features_full.h, line 14 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlySvyYOgoJOSo89DTv)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/features_full.h#L14)):</sup>
this was done to shorten composite guards from #if defined(foo) || defined(bar) to just #if foo || bar, to reduce "noise" when reading.

---

<sup>**[src/features_minimal.h, line 3 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyW_wxL1IQN2cestxU)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/features_minimal.h#L3)):</sup>
i want a prefix that is both short and apt but distinct (i.e. won't clash with manually created V7_ENABLE_FOO, of which we already have a few). having thought about it some more, i doubled the first underscore as well, so now it's V7_ENABLE__Foo__bar, and V7_ENABLE__Foo doesn't look as ugly as V7_ENABLE_Foo__ would.
added V7_ENABLE__Date which wraps entire date.c.

---

<sup>**[src/internal.h, line 81 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyQTYCuh7GIhyBtHhp)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/internal.h#L81)):</sup>
Done.

---

<sup>**[src/sources.mk, line 7 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyTC0FloTt3UIR8X-D)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/src/sources.mk#L7)):</sup>
i thought i did... done now.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

cpq:
Reviewed files:
- .gitignore @ r3
- Makefile @ r3
- scripts/gen-features-full.pl @ r3
- src/date.c @ r3
- src/features_full.h @ r3
- src/features_medium.h @ r3
- src/features_minimal.h @ r3
- src/internal.h @ r3
- src/sources.mk @ r3
- v7.c @ r3

---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

cpq:
LGTM

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

mmikulicic:
---

<sup>**[Makefile, line 40 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/446#-JlyURIcS-D3mQ2ee3qD)** ([raw file](https://github.com/cesanta/v7/blob/8c89317f224a26f9a474ad6c4dcf97e22c4ab2d5/Makefile#L40)):</sup>
SGTM

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

mmikulicic:
LGTM

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

mmikulicic:
Please rebase and solve conflicts (:drumroll:)

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

rojer:
done, merging to master now.

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/446)

PR #458, opened 2015-04-03
Pull out feature-related stuff into v7_features.h

rojer:
So it can be included into files that do not otherwise need internal.h.

Why not features.h? Because of /usr/include/features.h.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/458)

### DIFF
--- a/scripts/gen-features-full.pl
+++ b/scripts/gen-features-full.pl
@@ -21,5 +21,5 @@ print <<EOF
  */
 
 $feature_defs
-#endif
+#endif /* V7_BUILD_PROFILE == V7_BUILD_PROFILE_FULL */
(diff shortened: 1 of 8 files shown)


mmikulicic:
Reviewed files:
- scripts/gen-features-full.pl @ r1
- src/features_full.h @ r1
- src/features_medium.h @ r1
- src/features_minimal.h @ r1
- src/internal.h @ r1
- src/sources.mk @ r1
- src/v7_features.h @ r1
- v7.c @ r1

---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/458)

Issue #371, opened 2015-03-16
Make regex and utf8 support optional


cpq:
Add preprosessor conditionals `V7_DISABLE_REGEXP` and `V7_DISABLE_UTF8` - those might save a lot of space.

cpq:
Rojer, please take this over.
As part of your conditional compilation effort, make UTF8 support optional - I think the easiest way is to provide UTF8 stubs, e.g.

    #ifdef V7_DISABLE_UTF8
      implements a subset of utf8 functions that v7 uses using simple stubs
    #else
    #endif

rojer:
this should be it.
some ECMA tests are broken by this but this will be handled when specifying profiles.

PR #467, opened 2015-04-05
Make UTF support optional

rojer:
This breaks some Unicode-related ECMA tests but we'll deal with those
later, when specifying profiles.

#371

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/467)

### DIFF
--- a/src/features_full.h
+++ b/src/features_full.h
@@ -33,5 +33,6 @@
 #define V7_ENABLE__Math__sqrt 1
 #define V7_ENABLE__Math__tan 1
 #define V7_ENABLE__RegExp 1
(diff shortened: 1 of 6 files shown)


cpq:
Can't merge - is it rebased against master HEAD ?

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/467)

rojer:
sorry, forgot. should be ok now.

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/467)

PR #471, opened 2015-04-06
Squash the last warning when compiling V7

rojer:
[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/471)

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -128,6 +128,13 @@ V7_PRIVATE void *v7_to_pointer(val_t v) {
   return (void *) (uintptr_t)(v & 0xFFFFFFFFFFFFUL);
 }
 
+v7_cfunction_t v7_to_cfunction(val_t v) {
+  /* Implementation is identical to v7_to_pointer but is separate since
+   * object pointers are not directly convertible to function pointers
+   * according to ISO C and generates a warning in -Wpedantic mode. */
+  return (v7_cfunction_t)(v & 0xFFFFFFFFFFFFUL);
+}
+
 val_t v7_object_to_value(struct v7_object *o) {
   if (o == NULL) {
     return V7_NULL;
@@ -145,10 +152,6 @@ val_t v7_function_to_value(struct v7_function *o) {
 
 struct v7_function *v7_to_function(val_t v) {
   return (struct v7_function *) v7_to_pointer(v);
-}
-
-v7_cfunction_t v7_to_cfunction(val_t v) {
-  return (v7_cfunction_t) v7_to_pointer(v);
 }
 
 v7_val_t v7_create_cfunction(v7_cfunction_t f) {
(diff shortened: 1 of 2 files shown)


mmikulicic:
Reviewed files:
- src/vm.c @ r1
- v7.c @ r1

---

<sup>**[src/vm.c, line 135 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/471#-JmEj5pxOwmrRQh6QAE7)** ([raw file](https://github.com/cesanta/v7/blob/779fbd0df4716556d05180c181c929bdc1752464/src/vm.c#L135)):</sup>
BTW, `v7_to_pointer` has to be fixed to handle sign extension (e used to have 48-bit wide bitfield but it's not portable). Some platforms (e.g. solaris) have the stack addresses growing from the real top of the address space.

Wouldn't this solve this warning as well: 

```
return (v7_cfunction_t)(uintptr_t) v7_to_pointer(v);
```

?

(This way we only have to care about fixing sign extension in `v7_to_pointer`).

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/471)

mmikulicic:
---

<sup>**[src/vm.c, line 135 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/471#-JmEj5pxOwmrRQh6QAE7)** ([raw file](https://github.com/cesanta/v7/blob/779fbd0df4716556d05180c181c929bdc1752464/src/vm.c#L135)):</sup>
Actually, on a second thought, I think both are wrong.

There might be archs which use 64-bit function pointers and 32 bit "int pointers", i.e. `sizeof(v7_cfunction_t) > sizeof(uintptr_t)`

Is the cast to `(uintptr_t)` really necessary ?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/471)

rojer:
---

<sup>**[src/vm.c, line 135 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/471#-JmEj5pxOwmrRQh6QAE7)** ([raw file](https://github.com/cesanta/v7/blob/779fbd0df4716556d05180c181c929bdc1752464/src/vm.c#L135)):</sup>
ok, how about this?
i'm leaving sign extension change out for now, feel free to follow up.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/471)

mmikulicic:
Reviewed files:
- src/vm.c @ r2
- v7.c @ r2

---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/471)

mmikulicic:
Cool, thanks. Sign extension TODO (not very high prio right now)

---



---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/471)

PR #474, opened 2015-04-08
Make to_str return the number of bytes needed

### LINKED ISSUE #402: `v7_to_json` does not allocate a larger buffer if needed because `to_str` reports only number of bytes written, no number of bytes needed
v7 -e "eval('print(111111111111111111111111); print(2222222222222222222222);')"
1.11111111111111109247e+023
2.22222222222222216397e+021
null

v7 -e "eval('print(111111111111111111111111); print(2222222222222222222222); print(333333333333333333333333333);')"
Exec error [eval('print(111111111111111111111111); print(2222222222222222222222); print(333333333333333333333333333);')]: {"message":"parse error at at line 1 col 99: [ print(111111111111111111"}
undefined

gelraen:
It was returning the number of bytes actually written to the buffer,
which is not what other code expects.

Fixes #402

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/474)

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -113,8 +113,11 @@ int v7_main(int argc, char *argv[], void (*init_func)(struct v7 *)) {
 
   if (!show_ast) {
     char buf[2000];
-    v7_to_json(v7, res, buf, sizeof(buf));
-    printf("%s\n", buf);
+    char *s = v7_to_json(v7, res, buf, sizeof(buf));
+    printf("%s\n", s);
+    if (s != buf) {
+      free(s);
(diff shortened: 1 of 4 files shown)


mmikulicic:
Reviewed files:
- src/main.c @ r1
- src/vm.c @ r1
- tests/ecma_report.txt @ r1
- v7.c @ r1

---



---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/474)

PR #486, opened 2015-04-26
Remove .gitignore

cpq:
[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/486)


cpq:
Sync with the dev repo

mmikulicic:
Hi, this commit contains binaries and other stuff that should be first added in the `dev` repo and which will be synced here with the publish script.
BTW why do you want to remove .gitignore from the v7 repo?

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/486)

PR #492, opened 2015-08-14
to be compiled with 64bit VisualStudio 2010

koron:
When I try to compile with VisualStudio 2010 64bit version, this change is needed.


cpq:
Thank you. Pushed https://github.com/cesanta/v7/commit/0a867095a5e1ec217c71376700980b25c3ae7b2a.
Internally we're using unamalgamated sources that later gets concatenated into a single v7.c - that's why we don't merge directly.

PR #495, opened 2015-08-25
Remove static assert by using explicit size

edvakf:
Instead of asserting the size of "unsigned short", we can use "uint16_t".

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/495)


mmikulicic:
thanks! merged in c32f6b01047b480c46029c

Issue #494, opened 2015-08-24
Can't execute if binary stack size > 65535?


edvakf:
I prepared a file with 5040 lines of `print(1);` and ran with the v7 command. I got 5040 lines of 1 and an `undefined`. Then I added one more line of `print(1);` and I got just one line of output, `undefined`.

I noticed that this is something to do with the second and third bytes of the binary AST as well as the length of it. I couldn't find out the part of the code related to this.

```
$ ./v7/v7 -b ./5040.js | xxd -c 8 | head -n 2
0000000: 01ff f400 004f 000c  .....O..
0000008: 1405 7072 696e 7413  ..print.
$ ./v7/v7 -b ./5040.js | wc -c
65525
```

```
$ ./v7/v7 -b ./5041.js | xxd -c 8 | head -n 2
0000000: 0100 0100 004f 000c  .....O..
0000008: 1405 7072 696e 7413  ..print.
$./v7/v7 -b ./5041.js | wc -c
65538
```

Can someone explain to me what's happening here? Also, is it an expected behavior? Is there a way to have more statements than this?

I use Mac OS X Yosemite.

mmikulicic:
yeah, currently we support only 16-bit offsets in our AST. Our main target are small scripts so I deprioritized this. Anyway, it's a bug and it should be fixed.

edvakf:
Thanks for your reply. Could you point me to where to look for "we support only 16-bit offsets in our AST."?

mmikulicic:
The code that encodes the offset is here: https://github.com/cesanta/v7/blob/master/v7.c#L6276
and other places define the size of the AST nodes based on the fact that the offset size is 2 bytes.

edvakf:
Thanks. I'm gonna close this issue.

PR #499, opened 2015-08-26
Warn when trying to parse a large script

edvakf:
Refs https://github.com/cesanta/v7/issues/494

It would be nice to warn a user about large script because I had to spend time trying to figure out what was happening :(

I put the code here because I saw another warning. Ideally this check should be done in `ast_modify_skip`, but there is a performance trade off to do so.

I'm not sure if I should make another `v7_err` value (maybe V7_AST_TOO_LARGE), which seems more sensible approach.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/499)


mmikulicic:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[v7.c, line 9661 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/499#-JxeB5tukkBnzMICGMmm)** ([raw file](https://github.com/cesanta/v7/blob/424375c0e3a15d9e59b5f418ee51a594b0f93b5f/v7.c#L9661)):</sup>
perhaps it's better to make it an error.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

mmikulicic:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[v7.c, line 9661 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/499#-JxeB5tukkBnzMICGMmm)** ([raw file](https://github.com/cesanta/v7/blob/424375c0e3a15d9e59b5f418ee51a594b0f93b5f/v7.c#L9661)):</sup>
`V7_AST_TOO_LARGE` sounds good to me

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

edvakf:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[v7.c, line 9661 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/499#-JxeB5tukkBnzMICGMmm)** ([raw file](https://github.com/cesanta/v7/blob/424375c0e3a15d9e59b5f418ee51a594b0f93b5f/v7.c#L9661)):</sup>
Do you have any preference as to where to check for the error?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

mmikulicic:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[v7.c, line 9661 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/499#-JxeB5tukkBnzMICGMmm)** ([raw file](https://github.com/cesanta/v7/blob/424375c0e3a15d9e59b5f418ee51a594b0f93b5f/v7.c#L9661)):</sup>
the best place would be to check in modify_skip.
But it would be an overkill to return an error code and check all over the place, or even worse make non local returns (aka setjmp).

Perhaps we could just put a flag in the ast structure, set it in modify_skip and check it here (and return an error here).

The rationale for that is that they are only offsets relative to the current position, so the overall AST can be bigger than the maximum offset. However currently this doesn't help since the root node and stuff with potentially large bodies (SCRIPT, FUNCTION) have a skip offset that points to their end. I guess one in SCRIPT is useless since it will encompass the whole AST anyway; but let's do that in a second phase.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

edvakf:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[v7.c, line 9661 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/499#-JxeB5tukkBnzMICGMmm)** ([raw file](https://github.com/cesanta/v7/blob/424375c0e3a15d9e59b5f418ee51a594b0f93b5f/v7.c#L9661)):</sup>
OK. I pushed some commits. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

edvakf:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[tests/large_ast.js, line 302 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/499#-Jxh4cgS0hGnwBBs5usq)** ([raw file](https://github.com/cesanta/v7/blob/3d9cc5524e8db63e075b2a0db419e367e9927aef/tests/large_ast.js#L302)):</sup>
Sorry about the large file. This is the minimal failing example I could make.

---

<sup>**[v7.c, line 1381 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/499#-Jxh5icE1Zg4O5npCC9f)** ([raw file](https://github.com/cesanta/v7/blob/3d9cc5524e8db63e075b2a0db419e367e9927aef/v7.c#L1381)):</sup>
uint8_t would be smaller unless int is 8 bit, but I saw int being used for a boolean flag in other parts of this file.

---

<sup>**[v7.c, line 6445 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/499#-Jxh5Orj0tw1WbF9KY5C)** ([raw file](https://github.com/cesanta/v7/blob/3d9cc5524e8db63e075b2a0db419e367e9927aef/v7.c#L6445)):</sup>
I guess I have to do this in `ast_free` too, but is there a reason refcnt is not zeroed there?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

mmikulicic:
Thanks, looks good, just a couple minor comments.

---

Reviewed 3 of 4 files at r2.
Review status: 3 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[v7.c, line 0 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/499#-Jxh5Orj0tw1WbF9KY5C)** ([raw file](https://github.com/cesanta/v7/blob/3d9cc5524e8db63e075b2a0db419e367e9927aef/v7.c#L0)):</sup>
no reason, just a bug

---

<sup>**[v7.c, line 1381 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/499#-Jxh5icE1Zg4O5npCC9f)** ([raw file](https://github.com/cesanta/v7/blob/3d9cc5524e8db63e075b2a0db419e367e9927aef/v7.c#L1381)):</sup>
the compiler wouldn't make it smaller unless we pragma pack, which is not portable.

---

<sup>**[v7.c, line 6261 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/499#-Jxh8vPa1FwrrBYzGIkg)** ([raw file](https://github.com/cesanta/v7/blob/3d9cc5524e8db63e075b2a0db419e367e9927aef/v7.c#L6261)):</sup>
hmm, if `sizeof(ast_off_t) == sizeof(ast_skip_t)` this is always false. I guess some compilers will even warn about that.

We aim (though not always succeed) to be warning free on all compilers and C dialects.
I guess we should `#ifndef V7_LARGE_AST` here.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

edvakf:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[v7.c, line 6261 \[r2\]](https://reviewable.io:443/reviews/cesanta/v7/499#-Jxh8vPa1FwrrBYzGIkg)** ([raw file](https://github.com/cesanta/v7/blob/3d9cc5524e8db63e075b2a0db419e367e9927aef/v7.c#L6261)):</sup>
OK. I've added a commit do this check only when `V7_LARGE_AST` not set. We could also remove `has_overflow` flag, though it would introduce at least 3 more ifdef check...

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/499)

mmikulicic:
Unsolicited pong: I didn't forget about this, I'm improving some tooling we use to integrate external contributions into our workflow.

cpq:
@edvakf  - internally, our sources are split by many files, and publishing procedure does amalgamation. So any external contribution requires un-amalgamation

edvakf:
I see. The fix I needed was already merged, so I'm not in a rush.

mmikulicic:
published in bd764c04027b7a913895ce52dc7218eff8b8ec55

PR #510, opened 2015-10-18
Improved String.prototype.split() : added String-based support, fixed a couple of issues

dimonomid:
As the title suggests, if the separator given to `String.prototype.split()` is anything but RegExp, it is interpreted as String, and source string gets splitted by the plain string instead of RegExp. It works if v7 is built without RegExp support.

In addition, a couple of issues fixed, so that behavior of `String.prototype.split()` is now identical to that of browsers (Firefox, Chrome, Opera)

#### 1:

Accordingly to spec, if pattern does not match an empty string, then the empty substring at the end of the source string should be included in the resulting array:

  - `'123'.split(RegExp('.'))` should evaluate to `[ '', '', '', '' ]`, but previously it evaluated to `[ '', '', '' ]`.
  - `'123'.split(/2*/)` should evaluate to `['1', '3']`, but previously it evaluated to `['1', '', '3']`

#### 2:

Accordingly to spec, `''.split(/.*/)` should evaluate to `[]`, not `[ '' ]`, since given pattern matches an empty string.

Details in spec: http://www.ecma-international.org/ecma-262/5.1/#sec-15.5.4.14

#### 3:

Non-ASCII strings were not handled correctly: for example, `'абв'.split(RegExp(''))` should evaluate to `["а","б","в"]`, but previously it evaluated to `["�","�","�","�","�","�"]`

#### 4:

When the pattern contains capture groups, resulting array contained extra items:

  - `'123'.split(/(x)*/)` should evaluate to `["1",undefined,"2",undefined,"3"]`, but previously it evaluated to `["1",undefined,"2",undefined,"3",undefined]` (notice an extra trailing `undefined`)

#### 5:

Limit wasn't handled correctly. For example, `'123'.split(/(x)*/, 2)` should evaluate to `["1",undefined]`, but it evaluated to `["1",undefined,"2",undefined]`

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/510)

### DIFF
--- a/tests/unit_test.c
+++ b/tests/unit_test.c
@@ -299,6 +299,9 @@ static const char *test_stdlib(void) {
   ASSERT_EVAL_STR_EQ(v7, "m[0]", "a");
   ASSERT_EQ(eval(v7, &v, "m[1]"), V7_OK);
   ASSERT(v7_is_undefined(v));
+  ASSERT_EVAL_NUM_EQ(v7, "m = '123'.split(RegExp('.')); m.length", 4.0);
+  ASSERT_EVAL_NUM_EQ(v7, "m = ''.split(RegExp('')); m.length", 0.0);
+  ASSERT_EVAL_NUM_EQ(v7, "m = ''.split(RegExp('.')); m.length", 1.0);
   ASSERT_EVAL_NUM_EQ(v7, "m = 'aa bb cc'.split(' '); m.length", 3.0);
   ASSERT_EVAL_NUM_EQ(v7, "m = 'aa bb cc'.split(' ', 2); m.length", 2.0);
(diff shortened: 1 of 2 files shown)


mmikulicic:
sorry, I didn't forget, I'll review this tomorrow.

dimonomid:
Sergey asked to keep PRs containing just 1 commit, so, here we go: https://github.com/cesanta/v7/pull/512 . Closing this one.

mmikulicic:
We don't mind (actually encourage) squashing and force pushing one's
personal feature branch. So next time feel free to do that instead of
creating a new PR.

dimonomid:
Oh, ok, got it.

Issue #217, opened 2015-02-03
Hoisted function causes segfault


mmikulicic:
```js
(function() {return f; function f(){}})
```

mmikulicic:
fixed by @dimonomid in 01956ca7d692a4504473517789353fa62f67988

PR #511, opened 2015-10-19
Fixes issue #217: there was a bug in to_str() with function declarations

dimonomid:
`to_str()` wasn't expecting anything but `VAR_DECL`, although there might as well be a `FUNC_DECL`. Resolves issue #217.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/511)

### DIFF
--- a/v7.c
+++ b/v7.c
@@ -8108,7 +8108,8 @@ V7_PRIVATE int to_str(struct v7 *v7, val_t v, char *buf, size_t size,
           var_end = ast_get_skip(a, var, AST_END_SKIP);
           ast_move_to_children(a, &var);
           while (var < var_end) {
-            V7_CHECK(v7, ast_fetch_tag(a, &var) == AST_VAR_DECL);
+            enum ast_tag tag = ast_fetch_tag(a, &var);
+            V7_CHECK(v7, tag == AST_VAR_DECL || tag == AST_FUNC_DECL);
             name = ast_get_inlined_data(a, var, &name_len);
             ast_move_to_children(a, &var);
             ast_skip_tree(a, &var);


mmikulicic:
Hi! 

Looks good, good catch. Comments inline. Please amend and force push your branch when addressing the review comments.


---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[v7.c, line 8112 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/511#-K1-FYCkKGa4i-W2Wvsc)** ([raw file](https://github.com/cesanta/v7/blob/6cb0bcf57388d8d3a00d57f0bc0439f20077b8ca/v7.c#L8112)):</sup>
let's avoid the block and the var shadowing.

```
enum ast_tag vtag = ast_fetch_tag(a, &var);
...
```


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/511)

dimonomid:
Done. However, a question: why should we avoid the block? I mean, in general, not only in this particular case.

We don't need this variable anymore, and if we limit its scope, the compiler can (and often does) allocate a smaller stack frame. So we can save a bit of RAM, which is very expensive on embedded systems (at least on those I was working with).

And last but not least, it's just a nice hint for the programmer who reads the code that we don't need this variable anymore.

mmikulicic:
Can you show me an example and a name of a compiler that actually does this?

I saw stuff like this:

```
int foo(int a) {
#ifdef USE_BLOCKS
  {
#endif
    char buf[512];
    tmp(buf);
    if (*buf == 0)
      return a;
#ifdef USE_BLOCKS
  }
#endif
  return foo(a + 1) + 42;
}
```

Regardless of whether `USE_BLOCKS` is defined or not, it compiles down to:

```
_foo:                                   ## @foo
  .cfi_startproc
  pushq %rbp
  movq  %rsp, %rbp
  pushq %r14
  pushq %rbx
  subq  $528, %rsp              ## imm = 0x210
  movl  %edi, %ebx
  movq  ___stack_chk_guard@GOTPCREL(%rip), %r14
  movq  (%r14), %r14
  movq  %r14, -24(%rbp)
  leaq  -544(%rbp), %rdi
  callq _tmp
  cmpb  $0, -544(%rbp)
  je  LBB0_2
  incl  %ebx
  movl  %ebx, %edi
  callq _foo
  movl  %eax, %ebx
  addl  $42, %ebx
LBB0_2:
  cmpq  -24(%rbp), %r14
  jne LBB0_4
  movl  %ebx, %eax
  addq  $528, %rsp              ## <----- here
  popq  %rbx
  popq  %r14
  popq  %rbp
  retq
```

This was clang-700.0.72.

```
_foo:
LFB1:
  pushq %rbx
LCFI0:
  movl  %edi, %ebx
  subq  $512, %rsp
LCFI1:
  movq  %rsp, %rdi
  call  _tmp
  cmpb  $0, (%rsp)
  movl  %ebx, %eax
  je  L3
  leal  1(%rbx), %edi
  call  _foo
  addl  $42, %eax
L3:
  addq  $512, %rsp
LCFI2:
  popq  %rbx
```

This was x86_64-apple-darwin14.4.0-gcc-5.2.0. 


I've seen similar stuff with gcc-lx106.

----

Anyway, my rationale against unnecessary blocks is that it:

1. adds unnecessary nesting
2. encourages variable name shadowing

We do use blocks here and there in our code base. They make sense in large functions (unit tests, large switches), when you don't want too much distance between a declaration and it's use (damned ansi C).

dimonomid:
I must admit that it makes sense only with structs, as I see. Microchip's XC32 (gcc-based) for MIPS, `-Os` , the following code:
```
struct stest {
   int a;
   int b;
};

extern struct stest getstruct(void);

int testblocks(int a)
{
   int ret = 1;

#if USE_BLOCKS
   {
#endif
      struct stest s1 = getstruct();
      ret = ret * s1.a;
#if USE_BLOCKS
   }
#endif

#if USE_BLOCKS
   {
#endif
      struct stest s2 = getstruct();
      ret = ret * s2.b;
#if USE_BLOCKS
   }
#endif

   return ret;
}
```
USE_BLOCKS is 0:
```
163:                 int testblocks(int a)
164:                 {
9D0083D4  27BDFFD8   ADDIU SP, SP, -40    ## <---- here
9D0083DC  AFBF0024   SW RA, 36(SP)
165:                    int ret = 1;
166:                 
167:                 #if USE_BLOCKS
168:                    {
169:                 #endif
170:                       struct stest s1 = getstruct();
9D0083D8  27A40010   ADDIU A0, SP, 16
9D0083E0  0F402241   JAL getstruct
9D0083E4  AFB00020   SW S0, 32(SP)
171:                       ret = ret * s1.a;
172:                 #if USE_BLOCKS
173:                    }
174:                 #endif
175:                 
176:                 #if USE_BLOCKS
177:                    {
178:                 #endif
179:                       struct stest s2 = getstruct();
9D0083E8  27A40018   ADDIU A0, SP, 24
9D0083EC  0F402241   JAL getstruct
9D0083F0  8FB00010   LW S0, 16(SP)
180:                       ret = ret * s2.b;
181:                 #if USE_BLOCKS
182:                    }
183:                 #endif
184:                 
185:                    return ret;
186:                 }
9D0083F4  8FA2001C   LW V0, 28(SP)
9D0083F8  8FBF0024   LW RA, 36(SP)
9D0083FC  72021002   MUL V0, S0, V0
9D008400  8FB00020   LW S0, 32(SP)
9D008404  03E00008   JR RA
9D008408  27BD0028   ADDIU SP, SP, 40
```
USE_BLOCKS is 1:
```
163:                 int testblocks(int a)
164:                 {
9D0083D4  27BDFFE0   ADDIU SP, SP, -32    ## <---- here
9D0083DC  AFBF001C   SW RA, 28(SP)
165:                    int ret = 1;
166:                 
167:                 #if USE_BLOCKS
168:                    {
169:                 #endif
170:                       struct stest s1 = getstruct();
9D0083D8  27A40010   ADDIU A0, SP, 16
9D0083E0  0F402241   JAL getstruct
9D0083E4  AFB00018   SW S0, 24(SP)
171:                       ret = ret * s1.a;
172:                 #if USE_BLOCKS
173:                    }
174:                 #endif
175:                 
176:                 #if USE_BLOCKS
177:                    {
178:                 #endif
179:                       struct stest s2 = getstruct();
9D0083E8  27A40010   ADDIU A0, SP, 16
9D0083EC  0F402241   JAL getstruct
9D0083F0  8FB00010   LW S0, 16(SP)
180:                       ret = ret * s2.b;
181:                 #if USE_BLOCKS
182:                    }
183:                 #endif
184:                 
185:                    return ret;
186:                 }
9D0083F4  8FA20014   LW V0, 20(SP)
9D0083F8  8FBF001C   LW RA, 28(SP)
9D0083FC  72021002   MUL V0, S0, V0
9D008400  8FB00018   LW S0, 24(SP)
9D008404  03E00008   JR RA
9D008408  27BD0020   ADDIU SP, SP, 32
```
And I remember that I started to care about it after a story when I had to allocate multiple small structures on stack, and placing them into individual blocks saved me about 100 bytes of stack, which is significant. So, the habit of applying this approach with primitive values is not probably worthwhile.

mmikulicic:
It's good to know that it does indeed make a difference for some compilers. I wish gcc-l106x was that smart. Or perhaps it is under some specific circumstances I didn't yet find.

I don't see why a compiler cannot perform the same optimization even if there is no explicit block (the compiler can know when a variable is last used);

Anyway, this optimization can be done when useful. Critical parts of v7 w.r.t to stack consumption are all the parser functions and the `i_eval_...` functions of the interpreter. The latter is becoming less important now since there is ongoing work to switch to a bytecode interpreter.

mmikulicic:
merged in 01956ca7d692a4504473517789353fa62f67988

dimonomid:
Agreed with the fact that compiler could be smarter by performing this optimization without explicit blocks; and with primitives (such as int) it is smart enough: no matter whether we use explicit blocks, no extra stack space is consumed. But with structs, compiler seems more dumb.

PR #512, opened 2015-10-19
Improved String.prototype.split() : added String-based support, fixed a couple of issues

dimonomid:
As the title suggests, if the separator given to `String.prototype.split()` is anything but RegExp, it is interpreted as String, and source string gets splitted by the plain string instead of RegExp. It works if v7 is built without RegExp support.

In addition, a couple of issues fixed, so that behavior of `String.prototype.split()` is now identical to that of browsers (Firefox, Chrome, Opera)

#### 1:

Accordingly to spec, if pattern does not match an empty string, then the empty substring at the end of the source string should be included in the resulting array:

  - `'123'.split(RegExp('.'))` should evaluate to `[ '', '', '', '' ]`, but previously it evaluated to `[ '', '', '' ]`.
  - `'123'.split(/2*/)` should evaluate to `['1', '3']`, but previously it evaluated to `['1', '', '3']`

#### 2:

Accordingly to spec, `''.split(/.*/)` should evaluate to `[]`, not `[ '' ]`, since given pattern matches an empty string.

Details in spec: http://www.ecma-international.org/ecma-262/5.1/#sec-15.5.4.14

#### 3:

Non-ASCII strings were not handled correctly: for example, `'абв'.split(RegExp(''))` should evaluate to `["а","б","в"]`, but previously it evaluated to `["�","�","�","�","�","�"]`

#### 4:

When the pattern contains capture groups, resulting array contained extra items:

  - `'123'.split(/(x)*/)` should evaluate to `["1",undefined,"2",undefined,"3"]`, but previously it evaluated to `["1",undefined,"2",undefined,"3",undefined]` (notice an extra trailing `undefined`)

#### 5:

Limit wasn't handled correctly. For example, `'123'.split(/(x)*/, 2)` should evaluate to `["1",undefined]`, but it evaluated to `["1",undefined,"2",undefined]`

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/510)

### DIFF
--- a/tests/unit_test.c
+++ b/tests/unit_test.c
@@ -89,6 +89,50 @@ static int check_value(struct v7 *v7, val_t v, const char *str) {
   return res;
 }
 
+static int check_js_expr(struct v7 *v7, val_t v_actual,
+                         const char *expect_js_expr) {
+  int res = 1;
+  v7_val_t v_expect;
+  enum v7_err e;
+
+  v7_own(v7, &v_actual);
+
+  /* execute expected value */
+  e = v7_exec(v7, expect_js_expr, &v_expect);
+  v7_own(v7, &v_expect);
+
+  if (e != V7_OK) {
+    /* failed to execute expected value */
+    printf("Exec expected '%s' failed, err=%d\n", expect_js_expr, e);
+    res = 0;
+  } else {
+    /* now, stringify both values (actual and expected) and compare them */
+
+    char buf_actual[2048];
+    char *p_actual = v7_to_json(v7, v_actual, buf_actual, sizeof(buf_actual));
+
+    char buf_expect[2048];
+    char *p_expect = v7_to_json(v7, v_expect, buf_expect, sizeof(buf_expect));
+
+    if (strcmp(p_actual, p_expect) != 0) {
+      _strfail(p_actual, p_expect, -1);
+      res = 0;
+    }
+    if (p_actual != buf_actual) {
+      free(p_actual);
+    }
+
+    if (p_expect != buf_expect) {
+      free(p_expect);
+    }
+  }
+
+  v7_disown(v7, &v_expect);
+  v7_disown(v7, &v_actual);
+
+  return res;
+}
+
 static int check_num(struct v7 *v7, val_t v, double num) {
   int ret = isnan(num) ? isnan(v7_to_number(v)) : v7_to_number(v) == num;
   (void) v7;
@@ -155,6 +199,8 @@ static int test_if_expr(struct v7 *v7, const char *expr, int result) {
 
 #define ASSERT_EVAL_EQ(v7, js_expr, expected) \
   _ASSERT_EVAL_EQ(v7, js_expr, expected, check_value)
+#define ASSERT_EVAL_JS_EXPR_EQ(v7, js_expr, expected) \
+  _ASSERT_EVAL_EQ(v7, js_expr, expected, check_js_expr)
 #define ASSERT_EVAL_NUM_EQ(v7, js_expr, expected) \
   _ASSERT_EVAL_EQ(v7, js_expr, expected, check_num)
 #define ASSERT_EVAL_STR_EQ(v7, js_expr, expected) \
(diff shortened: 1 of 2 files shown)


mmikulicic:
Awesome! Thank you! 

A couple minor comments inline


---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit_test.c, line 92 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MTAA6QOodxXGsyt2)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L92)):</sup>
why not `check_json` ?


---

<sup>**[tests/unit_test.c, line 99 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MzqKY9l8KWHsJwKG)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L99)):</sup>
we'd like to make it build on ansi C as well; please don't mix declarations and code.


---

<sup>**[tests/unit_test.c, line 130 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MpQubsbrkLB4HXa_)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L130)):</sup>
```c
v7_disown(v7, &v_expect);
v7_disown(v7, &v_actual);
```


---

<sup>**[tests/unit_test.c, line 202 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MW4bwRPGEBIFEzWN)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L202)):</sup>
same here `ASSERT_EVAL_JSON_EQ`


---

<sup>**[v7.c, line 17576 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13Q6M0rNs_XHVCA6UH)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17576)):</sup>
BSD comments, see below


---

<sup>**[v7.c, line 17596 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13P5e-OJ0pNf0qxd6L)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17596)):</sup>
minor style:

BSD block comments

```c
/*
 * blah
 * blah
 */
```

for the rest of the styling we just defer to `clang-format`; it's ugly sometimes but at least we don't have to argue about style. Unfortunately it doesn't reformat comments :-)


---

<sup>**[v7.c, line 17609 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13PC4W6tzS8Evw_8fV)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17609)):</sup>
BSD block comments


---

<sup>**[v7.c, line 17635 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13QTt88md7oYy0PYrX)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17635)):</sup>
why don't we just ifdef this out, so that we don't need to have the dummy stub. I don't enjoy ifdef spaghetti either, but here we're just trading it with a different pasta brand .

Usually regexp will be disabled on platforms where every byte counts, so perhaps we can avoid adding no-op code if it doesn't greatly simplify the code. In this case I don't see how this indirection simplifies things. It seems to me that whenever a comment is necessary to specify that something happens only when something else happens, the code in question is not straightforward.


---

<sup>**[v7.c, line 17648 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13PIt5z-JF3Bid18ux)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17648)):</sup>
BSD block comments


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/512)

dimonomid:
Thanks for the review! I'll perform suggested changes tomorrow or maybe today's evening.

dimonomid:
Review status: 0 of 2 files reviewed at latest revision, 9 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit_test.c, line 92 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MTAA6QOodxXGsyt2)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L92)):</sup>
When I first read your comment, I thought I missed some ready-made function `check_json`, but I don't see such function. Do you mean that I'd better name the function `check_json` instead of `check_js_expr`?

I've chosen `check_js_expr` because the expression doesn't have to be a valid JSON: for example, single-quoted strings aren't a valid JSON. For JSON, instead of ` "['1', '2', '3']"`, we'd have to type ` "[\"1\",\"2\",\"3\"]"`, which is more tedious and less readable.

One more note: as I see, there is a function `check_value`, which checks the provided string against the exact output of `v7_to_json`. I decided not to use it either, since exact output of `v7_to_json` is an implementation detail: we should specify string exactly as it is formatted by `v7_to_json`: no extra spaces, etc. And if one day in the future `v7_to_json` is improved to provide string that is formatted a bit differently, tests that rely on `check_value` will probably be broken.

Please let me know if I missed something.


---

<sup>**[tests/unit_test.c, line 99 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MzqKY9l8KWHsJwKG)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L99)):</sup>
Sorry, my bad: currently, I don't have a habit to write ANSI C, I'll do my best to develop it.


---

<sup>**[tests/unit_test.c, line 130 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MpQubsbrkLB4HXa_)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L130)):</sup>
Ohh sorry, how could I miss that..
The appropriate compiler's warning were buried under lots of formatting warnings like this:

```
tests/unit_test.c:2334:3: warning: ISO C90 does not support the ‘%lf’ gnu_printf format [-Wformat=]
   RUN_TEST(test_gc_sweep);
   ^
tests/unit_test.c:2336:3: warning: ISO C90 does not support the ‘%lf’ gnu_printf format [-Wformat=]
   RUN_TEST(test_gc_own);
   ^
```

So I missed useful ones. Probably need to get rid of these formatting warnings. Thanks.


---

<sup>**[tests/unit_test.c, line 202 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MW4bwRPGEBIFEzWN)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L202)):</sup>
Answered above


---

<sup>**[v7.c, line 17576 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13Q6M0rNs_XHVCA6UH)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17576)):</sup>
Fixed


---

<sup>**[v7.c, line 17596 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13P5e-OJ0pNf0qxd6L)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17596)):</sup>
Ok! By the way, I like the suggested style more, but as I see, the code is generally written in a rather dense manner, so I tried not to bloat it with huge comment blocks. I got it: I tried too hard. :)

You've mentioned `clang-format`. So should I use it before publishing the code? I tried to feed `v7.c` to `clang-format-3.4`, it complained about non-existing `.clang-format` file and reformatted quite a few lines, so for now I haven't applied it. Do you use some custom `.clang-format` file?


---

<sup>**[v7.c, line 17609 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13PC4W6tzS8Evw_8fV)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17609)):</sup>
Fixed


---

<sup>**[v7.c, line 17635 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13QTt88md7oYy0PYrX)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17635)):</sup>
Agreed, avoiding a no-op and an extra pointer in the stack space for low-end devices sounds like a good idea, even though it brings some ifdef spaghetti.


---

<sup>**[v7.c, line 17648 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13PIt5z-JF3Bid18ux)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17648)):</sup>
Fixed


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/512)

mmikulicic:
LGTM, I'll merge it


---

Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[tests/unit_test.c, line 92 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13MTAA6QOodxXGsyt2)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/tests/unit_test.c#L92)):</sup>
you're right, `check_js_expr` is better; if we ever change the fine details of stringification it won't break the tests for nothing.


---

<sup>**[v7.c, line 17596 \[r1\]](https://reviewable.io:443/reviews/cesanta/v7/512#-K13P5e-OJ0pNf0qxd6L)** ([raw file](https://github.com/cesanta/v7/blob/471d06f656da2a2675f7bf874945505fdba4e35c/v7.c#L17596)):</sup>
yes, don't worry about it, I'll handle that at merge time. We have a convoluted internal process.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/512)

PR #513, opened 2015-10-22
Fixed String.prototype.lastIndexOf()

dimonomid:
At least, the following is fixed:

- `lastIndexOf()` always returned `-1` if `fromIndex` isn't provided
- if `fromIndex` is more than or equal to the string length, then `-1` was always returned, but we should search the whole string instead
- `'123456'.lastIndexOf('3456', 2)` returned `-1`, but it should return `2`

Added all appropriate unit tests as well.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/513)


mmikulicic:
LGTM


---

Reviewed 2 of 2 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/v7/513)

mmikulicic:
merged in 0c6fee3be8ec17faa635dfcabcdf72486a728347

PR #519, opened 2015-12-05
build: Support for VS2015

### LINKED ISSUE #493: Error building v7 on Windows 10 + VC10
timezone is _timezone and tzname is _tzname

Can make it work with this construct below but I guess it should be fixed in the beginning of the date block byt I am uncertain if this is just for Windows 10

#ifdef WIN32
  g_gmtoffms = _timezone * msPerSecond;
#else
  g_gmtoffms = timezone * msPerSecond;
#endif

  /*
   * tzname could be changed by localtime_r call,
   * so we have to store pointer
   * TODO(alashkin): need restart on tz change???
   */

#ifdef WIN32
  g_tzname = _tzname[0];
#else
  g_tzname = tzname[ 0 ];
#endif

jasonwilliams200OK:
Fixes #493

[Review on Reviewable](https://reviewable.io/reviews/cesanta/v7/519)

### DIFF
--- a/v7.c
+++ b/v7.c
@@ -50,16 +50,17 @@
 
 #define V7_VERSION "1.0"
 
-#if (defined(_WIN32) && !defined(__MINGW32__) && !defined(__MINGW64__)) || \
-    (defined(_MSC_VER) && _MSC_VER <= 1200)
-#define V7_WINDOWS
-#endif
+#ifdef _MSC_VER
 
-#ifdef V7_WINDOWS
+#if _MSC_VER >= 1900
+#define timezone _timezone
+#define tzname _tzname
+#endif
 typedef unsigned __int64 uint64_t;
 #else
 #include <inttypes.h>
 #endif
+
 typedef uint64_t v7_val_t;
 
 #ifdef __cplusplus
@@ -337,7 +338,7 @@ int v7_is_true(struct v7 *v7, v7_val_t v);
  *
  * Result can be NULL if you don't care about the return value.
  */
-enum v7_err v7_apply(struct v7 *, v7_val_t *result, v7_val_t func,
+enum v7_err v7_apply(struct v7 *, v7_val_t * volatile result, v7_val_t func,
                      v7_val_t this_obj, v7_val_t args);
 
 /*
@@ -1122,6 +1123,7 @@ char *utfutf(char *s1, char *s2);
 #endif
 #include <windows.h>
 #include <process.h>
+#include <errno.h>
 #ifndef EINPROGRESS
 #define EINPROGRESS WSAEINPROGRESS
 #endif
@@ -1133,9 +1135,7 @@ char *utfutf(char *s1, char *s2);
 #define STR(x) STRX(x)
 #define __func__ __FILE__ ":" STR(__LINE__)
 #endif
-#define snprintf _snprintf
 #define fileno _fileno
-#define vsnprintf _vsnprintf
 #define sleep(x) Sleep((x) *1000)
 #define to64(x) _atoi64(x)
 #define popen(x, y) _popen((x), (y))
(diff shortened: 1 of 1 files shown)


jasonwilliams200OK:
Tested on:

* VS2015 Update 1 on Windows 10 x64. Compiled for x86, amd64 and arm.
* V2013 on Windows 7 x86.

cpq:
Pushed https://github.com/cesanta/v7/commit/2593cb5fb03a5a00da462ab1514c2dcf71b45b5b

@jasonwilliams200OK please sync to head and amend your commit, keep only a single commit in the PR.
@dimonomid could you take care of this PR please?

jasonwilliams200OK:
@cpq, in case you missed it, i had only one commit in this PR. I didn't understood why did you publish 2593cb5?

cpq:
@jasonwilliams200OK Peter, the reason we do not integrate PRs directly is that public repo has source of V7. Our private repo is non-amalgamated, and that's where the integration really goes.
I've included your changes while you were still working on them, hence is it partial.

jasonwilliams200OK:
@cpq, thanks. Didn't knew about the private repo. :)
I have fixed up and made some additional warning fixes. Now with there are 0 build warning when compiling with VS2015 and 2013.

If i have a vote, I would say drop the VS support older than 2015. Consumers can download free [Build Tools 2015](http://www.microsoft.com/en-us/download/details.aspx?id=49983) (CLI compiler without VS), VS2015 Community, VS2015 Express for Desktop Editions (or higher) to compiler for XP-10 and WinRT targets as 2015 has quite sufficient support for C99 as well. This will clean up most of the `ifdefs _MSC_VER` and `_WIN32` blocks.

dimonomid:
@jasonwilliams200OK, could you please elaborate a bit more:

- I just tried to compile v7 with VC 19, and build fails both on our master and your PR because of the issue with `strnlen`: https://github.com/cesanta/v7/issues/523 . We'll resolve the issue in this repository soon (we'll not provide custom `strnlen` for VC starting from version 16), but I'm surprised by the fact that you aren't affected by the issue. If you have any ideas about it, I'd be glad to hear them.

- I see that you've removed some of the `const` qualifiers, just because VC generates a warning on the code like this:
```
      const char *p;
      /* ... */
      memcpy(&p, s + llen, sizeof(p));
```
But this is actually a VC's fault, the code is perfectly legal, and we're not going to remove `const` just in favour of VC.

- I see you removed a comment:
```
/*		
 * MSVC++ 14.0 _MSC_VER == 1900 (Visual Studio 2015)		
 * MSVC++ 12.0 _MSC_VER == 1800 (Visual Studio 2013)		
 * MSVC++ 11.0 _MSC_VER == 1700 (Visual Studio 2012)		
 * MSVC++ 10.0 _MSC_VER == 1600 (Visual Studio 2010)		
 * MSVC++ 9.0  _MSC_VER == 1500 (Visual Studio 2008)		
 * MSVC++ 8.0  _MSC_VER == 1400 (Visual Studio 2005)		
 * MSVC++ 7.1  _MSC_VER == 1310 (Visual Studio 2003)		
 * MSVC++ 7.0  _MSC_VER == 1300		
 * MSVC++ 6.0  _MSC_VER == 1200		
 * MSVC++ 5.0  _MSC_VER == 1100		
 */
```
Why? I find the comment quite useful.

jasonwilliams200OK:
@dimonomid, thanks for the review.

> I just tried to compile v7 with VC 19, and build fails both on our master and your PR because of the issue with strnlen: #523 . We'll resolve the issue in this repository soon (we'll not provide custom strnlen for VC starting from version 16), but I'm surprised by the fact that you aren't affected by the issue. If you have any ideas about it, I'd be glad to hear them.

I am compiling with Visual Studio 2015 on Windows 10 and VS 2013 on Windows 7 as follow:

```cmd
:: in cmd
"\Program Files (x86)\Microsoft Visual Studio 14.0\vc\vcvarsall.bat"

git clone https://github.com/jasonwilliams200ok/v7
cd v7
cl -c v7.c
lib v7.obj -OUT:v7.lib
```

I have 0 warnings and 0 errors.

> I see that you've removed some of the const qualifiers, just because VC generates a warning on the code like this:

Fair enough, I will bring them back. As a keen learner of C, I would love to understand and report a proper bug to VC team (at http://connect.microsoft.com/VisualStudio) about this, if that warning is really bogus. But looking at the flow and usage of the variable, it seems like we actually want a non-const `p` there.

> I see you removed a comment:

It was intentional. I will put that chunk back.

---

If you prefer, I can setup AppVeyor CI for this repo as part of this PR. It will take me very little effort to set it up and help future contributors to see status of build on Windows before the review starts.

dimonomid:
> But looking at the flow and usage of the variable, it seems like we actually want a non-const `p` there.

Not sure we understand each other, so, to make it clear: `p` **is** non-const here. The `const` applies to the data `p` points to, not to the `p` itself. And `memcpy` shouldn't care about it here, since we're writing to `&p`, not to `p`.

dimonomid:
As to AppVeyor: thank you, that sounds interesting, I'll discuss this with the team.

jasonwilliams200OK:
@dimonomid, I have fixed up the changes. Now the delta is bit smaller.

dimonomid:
@jasonwilliams200OK, for your information: it breaks VC6 build. We'll work on that.

jasonwilliams200OK:
@dimonomid, out of curiosity, what is the benefit of supporting two decade old VC6 (released 1998) if the target Windows SKUs are XP and above for ARM, MIPS, x86 and amd64? VS2013 and VS2015 do complies for the said targets.

dimonomid:
@jasonwilliams200OK, the benefit is not to drop support for people that still uses pre-C99 compilers and other old stuff. I'm not a fan of it either, but even I know even a major car alarm company which deliberately doesn't use C99 compiler for their embedded projects. I don't consider it as the very good thing to do, but it is what it is.

W.r.t. VC6 itself: it compiles for any version of Windows starting from Win95, and it doesn't need any external dependencies.

jasonwilliams200OK:
@dimonomid, I have tested building the latest master with VS2015 Update 2 and it builds with two warnings with `cl -c v7.c`:

```cmd
v7.c(17367): warning C4090: 'function': different 'const' qualifiers
v7.c(26955): warning C4028: formal parameter 4 different from declaration
```

With analyzer on: `cl -c v7.c /analyze` there are couple of warnings which I have logged here: https://gist.github.com/jasonwilliams200OK/c30010dfcab59ea2f70b0f1186c3c4bf. Some of them might be interesting.

Since the original issue is resolved, I am closing this.

Thanks for your help and insights. :)

PR #575, opened 2016-10-21
Update v7_main signature

brimstone:
I noticed the signature in the docs didn't match up with the code. I tried to fix this.

### DIFF
--- a/docs/c-api/main_public.h/v7_main.md
+++ b/docs/c-api/main_public.h/v7_main.md
@@ -3,7 +3,7 @@ title: "v7_main()"
 decl_name: "v7_main"
 symbol_kind: "func"
 signature: |
-  int v7_main(int argc, char *argv[], void (*pre_freeze_init);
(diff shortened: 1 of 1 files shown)


mmikulicic:
Hi! Thanks for helping. Unfortunately that's a generated markdown file. Our script is obviously confused by the fact we have an `)` inside the argument list.

We'll fix the script. Thanks!

brimstone:
Where's the script, want help fixing it?

mmikulicic:
unfortunately it's in a private repo; the script itself is not top secret or anything; but due to technical reasons it would be hard to opensource it;

but what if instead we work around this by also cleaning up the code a little bit:

`void (*pre_freeze_init)(struct v7 *)` -> `v7_init_cb_t pre_freeze_init`

and make a `typedef void (*v7_init_cb_t)(struct v7 *)` ?

What do you think?

brimstone:
I'm just starting to use the project, so I don't really have opinions yet. A callback type is probably fine.

